PR #4, opened 2026-02-05
PR Tracking Initial Release

donald-pinckney:
**Do NOT merge until good for releasing 0.1.0**

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: temporal-developer
 description: This skill should be used when the user asks to "create a Temporal workflow", "write a Temporal activity", "debug stuck workflow", "fix non-determinism error", "Temporal Python", "Temporal TypeScript", "Temporal Go", "Temporal Golang", "workflow replay", "activity timeout", "signal workflow", "query workflow", "worker not starting", "activity keeps retrying", "Temporal heartbeat", "continue-as-new", "child workflow", "saga pattern", "workflow versioning", "durable execution", "reliable distributed systems", or mentions Temporal SDK development.
-version: 1.0.0
+version: 0.1.0
 ---
 
 # Skill: temporal-developer


references/core/ai-integration.md
```diff
@@ -0,0 +1,237 @@
...
+| Debugging | Full execution history |
+
+## Core Patterns
+
+### Pattern 1: Generic LLM Activity
+
+Create flexible, reusable activities for LLM calls:
+
+```
+Activity: call_llm_generic(
```
cretz: What language format is this supposed to be, or is this pseudocode acceptable as kinda a made up format? I am struggling to review without knowing what things like `?` means in this unknown language

cretz: I can't tell, is this discouraging our pattern of putting all parameters into a single class (and having fields as optional)?

references/core/ai-integration.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Pattern 3: Centralized Retry Management
+
+**Critical**: Disable retries in LLM client libraries, let Temporal handle retries.
+
+```
+LLM Client Config:
+  max_retries = 0  ← Disable client retries
+
+Activity Retry Policy:
```
cretz: Is this encouraging specific retry policy or just saying that one is needed? It may accidentally be doing the former, at which point it should probably just leave the default policy (but can set max attempts if we don't want to discuss timeouts)

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+**Symptoms**:
+- Duplicate log entries
+- Multiple notifications for the same event
+- Inflated metrics
+
+**The Fix**:
+- Use the SDK's replay-aware logger (only logs on first execution)
+- Put all side effects in Activities
+
+### Non-Deterministic Time
```
cretz: Random too, IO too, threading too

references/core/gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+Local Activities skip the task queue for lower latency, but they're still subject to retries. The same idempotency rules apply.
+
+## Replay Safety Violations
+
+### Side Effects in Workflow Code
+
+**The Problem**: Code in workflow functions runs on first execution AND on every replay. Any side effect (logging, notifications, metrics) will happen multiple times.
+
+**Symptoms**:
```
cretz, line 23: Non-determinism errors here too

references/core/gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Multiple Workers with Different Code
+
+**The Problem**: If Worker A runs part of a workflow with code v1, then Worker B (with code v2) picks it up, replay may produce different Commands.
+
+**Symptoms**:
+- Non-determinism errors after deploying new code
+- Errors mentioning "command mismatch" or "unexpected command"
+
+**The Fix**:
```
cretz, line 48: Patching too

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+**The Problem**: Workflows started with old code continue running after you change the code.
+
+**Symptoms**:
+- Workflows behave unexpectedly after code changes
+- Non-determinism errors on previously-working workflows
+
+**The Fix**:
+- Terminate stale workflows: `temporal workflow terminate --workflow-id <id>`
+- Use `find-stalled-workflows.sh` to detect stuck workflows
+- In production, use versioning for backward compatibility
```
cretz: Or patching

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Issues**:
+- Hard to test and maintain
+- Event history grows unbounded
+- Single point of failure
+- Difficult to reason about
+
+**The Fix**:
+- Keep workflows focused on a single responsibility
+- Use Child Workflows for sub-processes
```
cretz: I disagree with this as an anti-pattern. In fact, IMO the anti-pattern is using child workflows as a code separation mechanism.

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Failing Too Quickly
+
+**The Problem**: Using aggressive retry policies that give up too easily.
+
+**Symptoms**:
+- Workflows failing on transient errors
+- Unnecessary workflow failures during brief outages
+
+**The Fix**: Use appropriate retry policies. Let Temporal handle transient failures with exponential backoff. Reserve `maximum_attempts=1` for truly non-retryable operations.
```
cretz: This should be noted that you are talking about activity retry policies here. We discourage workflow retry policies as an anti-pattern in most cases

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Query Handler Mistakes
+
+### Modifying State in Queries
```
cretz: Same applies to update validators

references/core/common-gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+**The Problem**: Queries are read-only. Modifying state in a query handler causes non-determinism on replay because queries don't generate history events.
+
+**Symptoms**:
+- State inconsistencies after workflow replay
+- Non-determinism errors
+
+**The Fix**: Queries must only read state. Use Updates for operations that need to modify state AND return a result.
+
+### Blocking in Queries
```
cretz: Same applies to update validators

references/core/gotchas.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Not Testing Replay Compatibility
+
+**The Problem**: Changing workflow code without verifying existing workflows can still replay.
+
+**Symptoms**:
+- Non-determinism errors after deployment
+- Stuck workflows that can't make progress
+
+**The Fix**: Use replay testing against saved histories from production or staging.
```
cretz, line 127: Another common fix is turning off workflow cache during non-Go testing, thereby forcing replays to occur. Not sure if we should suggest this, would like to hear others' opinions.

references/core/determinism.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Overview
+
+Temporal workflows must be deterministic because of **history replay** - the mechanism that enables durable execution.
+
+## Why Determinism Matters
+
+### The Replay Mechanism
+
+When a Worker needs to restore workflow state (after crash, cache eviction, or continuing after a long timer), it **re-executes the workflow code from the beginning**. But instead of re-running activities, it uses results stored in the Event History.
```
cretz:
```suggestion
When a Worker needs to restore workflow state (after crash, cache eviction, or continuing after a long timer), it **re-executes the workflow code from the beginning**. But instead of re-running external actions, it uses results stored in the Event History.
```

references/core/determinism.md
```diff
@@ -0,0 +1,144 @@
...
+
+Replay (Recovery):
+  Code runs again → Generates Commands → SDK compares to Events
+  If match: Use stored results, continue
+  If mismatch: NondeterminismError!
+```
+
+### Commands and Events
+
+Every workflow operation generates a Command that becomes an Event:
```
cretz: This is non-exhaustive

```suggestion
Every workflow operation generates a Command that becomes an Event, here are some examples:
```

references/core/determinism.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Non-Deterministic Iteration
+- Map/dict iteration order (in some languages)
+- Set iteration order
+
+### Threading/Concurrency
+- Race conditions produce different outcomes
+- Non-deterministic ordering
+
+## SDK Protection Mechanisms
```
cretz, line 78: Should also list Go `workflowcheck` static analyzer, Java `workflowcheck` static analyzer, .NET tracing event listener (catches threading issues only), and Ruby's `TracePoint`-based method checker

references/core/error-reference.md
cretz, line 1: We have some rules numbers that come up on some of these errors, ref https://github.com/temporalio/rules

references/core/interactive-workflows.md
```diff
@@ -0,0 +1,53 @@
...
+uv run interact --workflow-id <id> --signal-name "signal_name" --data '{"key": "value"}'
+```
+
+## Updates
+
+Request-response style interaction (returns a value).
+
+```bash
+# Send update to workflow
+temporal workflow update \
```
cretz: Invalid CLI call

references/core/interactive-workflows.md
```diff
@@ -0,0 +1,53 @@
...
+Read-only inspection of workflow state.
+
+```bash
+# Query workflow state (read-only)
+temporal workflow query \
+  --workflow-id <id> \
+  --name "get_status"
+```
+
+## Testing Interactive Workflows
```
cretz: This is a strange section full of platform-specific things not defined anywhere. Testing interactive workflows should be done as you would test any other workflow

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+
+**Purpose**: Read workflow state synchronously without modifying it.
+
+**When to Use**:
+- Building dashboards showing workflow progress
+- Health checks and monitoring
+- Debugging workflow state
+- Exposing current status to external systems
+
+**Characteristics**:
```
cretz, line 49: * Can run even on completed workflows

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+**When to Use**:
+- Operations that need confirmation (add item, return count)
+- Validation before accepting changes
+- Replace signal+query combinations
+- Request-response patterns within workflow
+
+**Characteristics**:
+- Synchronous - caller waits for completion
+- Can mutate state AND return values
+- Supports validators to reject invalid updates
```
cretz:
```suggestion
- Supports validators to reject invalid updates before they even get persisted into history
```

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+Client                    Workflow
+  │                          │
+  │──── update(addItem) ────▶│
+  │                          │ (validates, modifies state)
+  │◀──── {count: 5} ─────────│
+```
+
+## Child Workflows
+
+**Purpose**: Break complex workflows into smaller, reusable pieces.
```
cretz: The other "when to use" bullets are more of a primary purpose than this IMO. This is actually an anti-pattern IMO.

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+
+**When to Use**:
+- Prevent history from growing too large
+- Isolate failure domains (child can fail without failing parent)
+- Reuse workflow logic across multiple parents
```
cretz: Not a big fan of this, programming languages have good code sharing mechanisms already, should not reach for child workflows as a code sharing mechanism unless one of the other benefits is true IMO.

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+**Purpose**: Run multiple independent operations concurrently.
+
+**When to Use**:
+- Processing multiple items that don't depend on each other
+- Calling multiple APIs simultaneously
+- Fan-out/fan-in patterns
+- Reducing total workflow duration
+
+**Patterns**:
+- `Promise.all()` / `asyncio.gather()` - Wait for all
```
cretz: After initial skill released, would like to discuss the rest of the langs that are not included here

cretz:
```suggestion
- `Promise` / `asyncio` - Use traditional concurrency helpers (e.g. wait for all, wait for first, etc)
```

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+**Use Cases**:
+- Scheduled reminders
+- Timeout handling
+- Delayed actions
+- Polling with intervals
+
+**Characteristics**:
+- Timers are durable (persisted in history)
+- Can be cancelled
+- Combine with cancellation scopes for timeouts
```
cretz: Cancellation scopes doesn't make sense in many SDKs

references/core/patterns.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Polling Pattern
```
cretz: There are actually 3 polling patterns (see samples in the SDKs), and for all but the most infrequent, this is actually the wrong one described here

references/core/tool-reference.md
cretz, line 1: If we want these kinds of scripts, IMO we need to develop them in a proper programming language (e.g. python) as opposed to shell. But what I would hope (and have seen), is that Claude can understand the CLI or SDK to do these things itself live. If we _must_ offer scripts, IMO we should be testing them. What does `uv run worker` even mean?

EDIT: Reading this whole file, it's very lang specific, uses very opaque approaches that are not accurate, and in general has all kinds of concerns. IMO we should completely get rid of these scripts in this iteration and rethink them. In many cases, Claude doesn't need help running processes.

references/core/tool-reference.md
```diff
@@ -0,0 +1,84 @@
...
+|------|-------------|-------------|
+| `analyze-workflow-error.sh` | Extract errors from history | `--workflow-id`, `--run-id` |
+| `get-workflow-result.sh` | Get workflow output | `--workflow-id`, `--raw` |
+| `bulk-cancel-workflows.sh` | Mass cancellation | `--pattern "..."` |
+
+## Worker Management Details
+
+### The Golden Rule
+
+**Ensure no old workers are running.** Stale workers with outdated code cause:
```
cretz: Note, this is not true in some situations where it can be important to run old workers for older workflows. Unsure if that applies to local dev much though.

references/core/tool-reference.md
```diff
@@ -0,0 +1,84 @@
...
+- Executing old buggy code
+- Confusing behavior
+
+**Best practice**: Run only ONE worker instance with the latest code.
+
+### Starting Workers
+
+```bash
+# PREFERRED: Smart restart (kills old, starts fresh)
+./scripts/ensure-worker.sh
```
cretz: This type of stuff is no good for anything but Python as I read the code. I think we should consider scrubbing this entire script concept and/or re-thinking it.

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+│   └─▶ COMPLETED (but wrong result)
+│       └─▶ Go to: "Wrong Result" section
+```
+
+## Workflow Stuck (RUNNING but No Progress)
+
+### Decision Tree
+
+```
+Workflow stuck in RUNNING?
```
cretz, line 28: This needs to include looking for a task failure. See https://docs.temporal.io/encyclopedia/detecting-workflow-failures#detecting-workflow-task-failures, though while that is rolling out, can just read history to see task failures

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Decision Tree
+
+```
+NondeterminismError?
+│
+├─▶ Was code intentionally changed?
+│   │
+│   ├─▶ YES: Use patching API
+│   │   └─▶ See: references/core/versioning.md
```
cretz: Or terminate or reset

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+│   │
+│   └─▶ NO: Accidental change
+│       │
+│       ├─▶ Can you identify the change?
+│       │   │
+│       │   ├─▶ YES: Revert and restart worker
```
cretz: This doesn't always work if workflow has progressed past the change

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+│       │   │
+│       │   └─▶ NO: Compare current code to expected history
+│       │       └─▶ Check: Activity names, order, parameters
+```
+
+### Common Causes
+
+1. **Changed activity order**
```
cretz: Should replace the word "activity" in these three bullets to "call". It applies to timers, child workflows, Nexus calls, etc.

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+   ```
+   # Before                    # After (BREAKS)
+   await process_order(...)    await handle_order(...)
+   ```
+
+3. **Added/removed activity call**
+   - Adding new activity mid-workflow
+   - Removing activity that was previously called
+
+4. **Using non-deterministic values**
```
cretz:
```suggestion
4. **Using non-deterministic code**
```

It's not just about values, it's about general non-deterministic code (e.g. threading)

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+```
+Workflow status = FAILED?
+│
+├─▶ Check workflow error message
+│   │
+│   ├─▶ Application error (your code)
+│   │   └─▶ Fix the bug, start new workflow
+│   │
+│   ├─▶ NondeterminismError
+│   │   └─▶ Go to: "Non-Determinism" section
```
cretz, line 155: Workflows don't fail for this reason unless this is opted into, which is a good idea for tests. We should consider noting how to opt-in to this somewhere.

references/core/troubleshooting.md
```diff
@@ -0,0 +1,308 @@
...
+│   │
+│   └─▶ Timeout error
+│       └─▶ Go to: "Timeout Issues" section
```
cretz, line 158: We discourage setting timeouts on workflows, unsure if we say that anywhere, but if not, we should

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+await schedule.unpause()
+await schedule.trigger()  # Run immediately
+await schedule.delete()
+```
+
+## Interceptors
+
+Interceptors allow cross-cutting concerns like logging, metrics, and auth.
+
+### Creating a Custom Activity Interceptor
```
cretz: It's rare to create activity and workflow interceptors independently, would suggest not encouraging it. This is also missing client interceptors and Nexus interceptors. This is also missing discussion of plugins which is a good way to configure interceptors.

references/core/versioning.md
```diff
@@ -0,0 +1,172 @@
...
+Original Code (recorded in history):
+  await activity_a()
+  await activity_b()
+
+Updated Code (during replay):
+  await activity_a()
+  await activity_c()  ← Different! NondeterminismError
+```
+
+## Approach 1: Patching API
```
cretz, line 27: Consider clarifying the `TemporalChangeVersion` search attribute which has a lot of value telling AI about

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+
+class LoggingInterceptor(Interceptor):
+    def workflow_interceptor_class(
+        self,
+        input: WorkflowInterceptorClassInput,
+    ) -> type[WorkflowInboundInterceptor] | None:
+        return LoggingWorkflowInboundInterceptor
+```
+
+## Dynamic Workflows and Activities
```
cretz: We should discourage this pattern. It's primary use was to get good activity/workflow names in UI, but now with summary the caller can set, I think we should just remove this section altogether (i.e. it's _too_ advanced)

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+
+```python
+@activity.defn(dynamic=True)
+async def dynamic_activity(args: Sequence[RawValue]) -> Any:
+    activity_type = activity.info().activity_type
+    # Handle based on type
+    ...
+```
+
+## Async Activity Completion
```
cretz, line 39: Need some discussion of heartbeat expectations (or lack of them) here. Also, somewhere in the general skills area, we need to discuss why one may choose signals over async activity (ref https://docs.temporal.io/activity-execution#when-to-use-async-completion)

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+        )
+```
+
+## Sandbox Customization
+
+The Python SDK runs workflows in a sandbox to ensure determinism. You can customize sandbox restrictions when needed.
+
+### Passing Through Modules
+
+If you need to use modules that are blocked by the sandbox:
```
cretz: I'd phrase it as we do in the README:

> For performance and behavior reasons, users are encouraged to pass through all modules whose calls will be deterministic.

Which is basically all of them. We of course automatically pass through all stdlib and `temporalio`

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+### Temporary Passthrough in Workflow Code
+
+```python
+@workflow.run
+async def run(self) -> str:
+    # Temporarily disable sandbox restrictions for imports
+    with workflow.unsafe.imports_passed_through():
+        import some_restricted_module
```
cretz: This is an anti-pattern, all workflow imports should be at the top of the file, not at runtime (even when passed through)

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+
+    if approved:
+        await client.get_async_activity_handle(task_token).complete("approved")
+    else:
+        await client.get_async_activity_handle(task_token).fail(
+            ApplicationError("Rejected")
+        )
+```
+
+## Sandbox Customization
```
cretz, line 79: We should discuss `import_notification_policy` and `disable_lazy_sys_module_passthrough` probably, even if just short blurbs. Specifically should probably encourage `import_notification_policy=SandboxImportNotificationPolicy.WARN_ON_DYNAMIC_IMPORT | SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH`. The first flag is default, but the second for warning on unintentional pass through is not for compatibility reasons, but has benefit for users if they are concerned about missing passthroughs.

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+```python
+@workflow.run
+async def run(self) -> str:
+    # Temporarily disable sandbox restrictions for imports
+    with workflow.unsafe.imports_passed_through():
+        import some_restricted_module
+        # Use the module...
+```
+
+### Customizing Invalid Module Members
```
cretz: There is also `workflow.unsafe.sandbox_unrestricted()` context manager that can be used inline

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+    # Executor for sync activities
+    activity_executor=ThreadPoolExecutor(max_workers=50),
+    # Graceful shutdown timeout
+    graceful_shutdown_timeout=timedelta(seconds=30),
+)
+```
+
+## Workflow Init Decorator
+
+Use `@workflow.init` to run initialization code when a workflow is first created (not on replay).
```
cretz:
```suggestion
Use `@workflow.init` to run initialization code when a workflow is first created.
```

Can clarify the purpose is to run before signal/update or run is invoked (because remember, signal/update can run _before_ run, and AI needs to know this and not assume run is first)

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+        self._value = initial_value
+        self._items: list[str] = []
+
+    @workflow.run
+    async def run(self) -> str:
+        # self._value and self._items are already initialized
+        return self._value
+```
+
+## Workflow Failure Exception Types
```
cretz, line 136: Also configurable at the worker level. Also, mentioned in previous comment, there is value in just setting this to `Exception` in tests

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+                system_prompt="You are a helpful assistant.",
+                user_input=prompt,
+            ),
+            start_to_close_timeout=timedelta(seconds=30),
+            retry_policy=RetryPolicy(
+                initial_interval=timedelta(seconds=1),
+                backoff_coefficient=2.0,
+                maximum_interval=timedelta(seconds=60),
+                maximum_attempts=5,
+                non_retryable_error_types=["InvalidAPIKeyError"],
```
cretz: We shouldn't make a caller set this every time, we should adjust the activity code to raise it as non-retryable. This is a general practice we tell people in the core level, we should follow here too IMO.

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+            LLMRequest(
+                model="gpt-4",
+                system_prompt="You are a helpful assistant.",
+                user_input=prompt,
+            ),
+            start_to_close_timeout=timedelta(seconds=30),
+            retry_policy=RetryPolicy(
+                initial_interval=timedelta(seconds=1),
+                backoff_coefficient=2.0,
+                maximum_interval=timedelta(seconds=60),
```
cretz: I'd leave these three off. The first two are already the defaults, the third is defaulted as 100x initial which is good enough. We should discourage altering retry policy values unless needed.

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+
+with workflow.unsafe.imports_passed_through():
+    from activities.llm import call_llm, LLMRequest, LLMResponse
+    from activities.tools import execute_tool
+    from models.tools import ToolDefinition
+
+@workflow.defn
+class AgentWorkflow:
+    @workflow.run
+    async def run(self, user_request: str, tools: list[ToolDefinition]) -> str:
```
cretz: We should not encourage multi-param workflows

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+        messages = []
+
+        while True:
+            # Phase 1: Get LLM response with tools
+            response = await workflow.execute_activity(
```
cretz, line 185: Inconsistent with whether we encourage max attempts on these calls

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+    try:
+        response = await openai_client.chat.completions.create(...)
+        return LLMResponse(...)
+    except openai.RateLimitError as e:
+        # Extract retry-after if available
+        retry_after = e.response.headers.get("retry-after", 30)
+        raise ApplicationError(
+            f"Rate limited, retry after {retry_after}s",
+            non_retryable=False,  # Allow Temporal to retry
+        )
```
cretz: Firstly, `non_retryable=False` is the default. Secondly, this should use the `next_retry_delay` option. Thirdly, users may want to set this as a benign exception which means it won't appear as a failure in logs/metrics (but maybe they don't).

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+            workflow.execute_activity(
+                search_web,
+                query,
+                start_to_close_timeout=timedelta(seconds=300),
+            )
+            for query in queries
+        ]
+
+        # Continue with partial results on failure
+        results = await asyncio.gather(*search_tasks, return_exceptions=True)
```
cretz, line 289: Consider timing this out (or using `schedule_to_close_timeout` on the activities) if you don't want one repeatedly failing to hang up the rest.

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+## OpenAI Agents SDK Integration
+
+Using Temporal's OpenAI contrib module:
+
+```python
+from temporalio.contrib.openai import create_workflow_agent
+from agents import Agent, Runner
+
+# Create a Temporal-aware agent
+agent = create_workflow_agent(
```
cretz: We should not encourage the use of global variables like this, should be created in the workflow

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+## Testing with Mocks
+
+```python
+import pytest
+from temporalio.testing import WorkflowEnvironment
+from temporalio.worker import Worker
+
+@pytest.fixture
+async def workflow_environment():
+    async with await WorkflowEnvironment.start_time_skipping() as env:
```
cretz: We should discourage the time-skipping environment in favor of the local environment unless time-skipping is needed as part of the test

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+        yield env
+
+async def test_llm_workflow(workflow_environment):
+    # Mock LLM activity
+    async def mock_call_llm(request):
```
cretz: Has this code been tested? Activities have to have a decorator, and I admit I haven't tested whether it works in a nested context

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+import pytest
+from temporalio.testing import WorkflowEnvironment
+from temporalio.worker import Worker
+
+@pytest.fixture
+async def workflow_environment():
+    async with await WorkflowEnvironment.start_time_skipping() as env:
+        yield env
+
+async def test_llm_workflow(workflow_environment):
```
cretz: We should encourage type hinting the fixture params

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+    async def mock_call_llm(request):
+        return LLMResponse(
+            content="Mocked response",
+            tool_calls=None,
+            usage={"total_tokens": 100},
+        )
+
+    async with Worker(
+        workflow_environment.client,
+        task_queue="test-queue",
```
cretz: We should discourage hard-coding task queues in tests, especially with names like this, it is dangerous for test isolation (recommend UUID)

references/python/ai-patterns.md
```diff
@@ -0,0 +1,362 @@
...
+        workflows=[LLMWorkflow],
+        activities=[mock_call_llm],  # Use mock
+    ):
+        result = await workflow_environment.client.execute_workflow(
+            LLMWorkflow.run,
+            "test prompt",
+            id="test-workflow",
```
cretz: We should discourage hard-coding workflow IDs in tests, especially with names like this, it is dangerous for test isolation (recommend UUID)

references/python/data-handling.md
```diff
@@ -0,0 +1,230 @@
...
+    total: float
+    customer_email: str
+
+class OrderResult(BaseModel):
+    order_id: str
+    status: str
+    tracking_number: str | None = None
+
+@workflow.defn
+class OrderWorkflow:
```
cretz, line 36: We should never put a workflow in the same file as non-workflow code, even in snippets, it is a dangerous practice and we've already found AI doesn't understand this well

references/python/data-handling.md
```diff
@@ -0,0 +1,230 @@
...
+        )
+
+# Configure client with Pydantic support
+client = await Client.connect(
+    "localhost:7233",
+    data_converter=pydantic_data_converter,
+)
+```
+
+## Custom Data Converter
```
cretz: This is often not the best way to implement custom conversion. Rather a specific encoding converter can often be clearer. See https://github.com/temporalio/samples-python/tree/main/custom_converter.

references/python/data-handling.md
```diff
@@ -0,0 +1,230 @@
...
+
+class EncryptionCodec(PayloadCodec):
+    def __init__(self, key: bytes):
+        self._fernet = Fernet(key)
+
+    async def encode(self, payloads: Sequence[Payload]) -> list[Payload]:
+        return [
+            Payload(
+                metadata={"encoding": b"binary/encrypted"},
+                data=self._fernet.encrypt(p.SerializeToString()),
```
cretz: This is blocking, and while encryption is CPU bound, it does use C extensions that give up the GIL, so no need to block async event loop here. So no blocking code should be in `async def` in Python ever.

```suggestion
                data=await asyncio.to_thread(self._fernet.encrypt, p.SerializeToString()),
```

This is even touched on in the [Python docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread):

> Note
> 
> Due to the [GIL](https://docs.python.org/3/glossary.html#term-GIL), asyncio.to_thread() can typically only be used to make IO-bound functions non-blocking. However, for extension modules that release the GIL or alternative Python implementations that don’t have one, asyncio.to_thread() can also be used for CPU-bound functions.

Same for decrypt below. Could show advanced form using thread pool executor, but the default executor on the event loop is good enough for demo.

references/python/advanced-features.md
```diff
@@ -0,0 +1,378 @@
...
+
+    # Store task token for later completion (e.g., in database)
+    await store_task_token(request_id, task_token)
+
+    # Raise to indicate async completion
+    activity.raise_complete_async()
+
+# Later, complete the activity from another process
+async def complete_approval(request_id: str, approved: bool):
+    client = await Client.connect("localhost:7233")
```
cretz: All of these samples are using a localhost target and leaving namespace as `"default"`. We cannot unset this default for compatibility reasons, but if we could, we'd force namespace to be set in `connect` (and we do in newer SDKs). I think we should discourage using `.connect` without `namespace=`.

references/python/data-handling.md
```diff
@@ -0,0 +1,230 @@
...
+    ),
+)
+```
+
+## Search Attributes
+
+Custom searchable fields for workflow visibility.
+
+```python
+from temporalio.common import SearchAttributes, SearchAttributeKey
```
cretz: I think this snippet confuses typed search attributes and the deprecated search attributes. This should use `TypedSearchAttributes`. `SearchAttributeKey` is meant for typed search attributes.

references/python/data-handling.md
```diff
@@ -0,0 +1,230 @@
...
+    },
+)
+
+# Read memo from workflow
+@workflow.defn
+class OrderWorkflow:
+    @workflow.run
+    async def run(self, order: Order) -> str:
+        memo = workflow.memo()
+        notes = memo.get("notes", "")
```
cretz: Would encourage `workflow.memo_value` which can accept a type hint

references/python/determinism.md
```diff
@@ -0,0 +1,143 @@
...
+### Example: Why datetime.now() Breaks Replay
+
+```python
+# BAD - Non-deterministic
+@workflow.defn
+class BadWorkflow:
+    @workflow.run
+    async def run(self) -> str:
+        import datetime
+        if datetime.datetime.now().hour < 12:  # Different value on replay!
```
cretz: Note, you can't even run this on Python today because our sandbox prevents this call today and gives a different error

references/python/determinism.md
```diff
@@ -0,0 +1,143 @@
...
+- Passes through standard library with restrictions
+
+## Safe Alternatives
+
+| Forbidden | Safe Alternative |
+|-----------|------------------|
+| `datetime.now()` | `workflow.now()` |
+| `datetime.utcnow()` | `workflow.now()` |
+| `random.random()` | `workflow.random().random()` |
+| `random.randint()` | `workflow.random().randint()` |
```
cretz: Note, we now offer `new_random()` and should probably encourage it

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+
+```python
+from temporalio.common import RetryPolicy
+
+result = await workflow.execute_activity(
+    my_activity,
+    schedule_to_close_timeout=timedelta(minutes=10),
+    retry_policy=RetryPolicy(
+        initial_interval=timedelta(seconds=1),
+        backoff_coefficient=2.0,
```
cretz: These two values are default, we should not encourage explicitly providing defaults for retry policies

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+
+```python
+from temporalio.exceptions import ActivityError
+
+@workflow.run
+async def run(self) -> str:
+    try:
+        return await workflow.execute_activity(
+            risky_activity,
+            schedule_to_close_timeout=timedelta(minutes=5),
```
cretz: Lots of samples here use `schedule_to_close_timeout` even when timeout unrelated, but we encourage `start_to_close_timeout` IIRC. Would suggest always using the latter unless there's a specific reason to use the former.

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
+# Python SDK Error Handling
+
+## Overview
+
+The Python SDK uses `ApplicationError` for application-specific errors and provides comprehensive retry policy configuration.
+
+## Application Errors
+
+```python
+from temporalio.exceptions import ApplicationError
```
cretz, line 11: This file has a lot of weird inconsistencies with imports. Sometimes it imports, sometimes not, and even when it does, it only imports some of the things.

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+        # Handle or re-raise
+        raise ApplicationError("Workflow failed due to activity error")
+```
+
+## Retry Policy Configuration
+
+```python
+from temporalio.common import RetryPolicy
+
+result = await workflow.execute_activity(
```
cretz: This file has a bunch of weird inconsistencies with whether regular Python code is in or out of a `run` function

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+## Workflow Failure
+
+```python
+@workflow.run
+async def run(self) -> str:
+    if some_condition:
+        raise ApplicationError(
+            "Cannot process order",
+            type="BusinessError",
+            non_retryable=True,
```
cretz: Workflows are not retried by default, so this doesn't need to be set (and we discourage opting in to workflow retry)

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Idempotency Patterns
+
+When Activities interact with external systems, making them idempotent ensures correctness during retries and replay.
+
+### Using Workflow IDs as Idempotency Keys
+
+```python
+@activity.defn
+async def charge_payment(order_id: str, amount: float) -> str:
```
cretz: Mentioned elsewhere, but will mention again, we should not encourage multi-param constructs in favor of single-param

references/python/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+3. **Make operations repeatable**: Ensure calling twice produces the same result
+4. **Record outcomes**: Store transaction IDs or results for verification
+
+## Best Practices
+
+1. Use specific error types for different failure modes
+2. Mark permanent failures as non-retryable
+3. Configure appropriate retry policies per activity
+4. Log errors before re-raising
+5. Use `ActivityError` to catch activity failures in workflows
```
cretz, line 137: It is strange to talk about activities like this but not talk about child workflows or Nexus operations. All of these skill pieces are extremely inconsistent with themselves

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+```
+
+### Blocking in Queries
+
+```python
+# BAD - Queries cannot await
+@workflow.query
+async def get_data_with_refresh(self) -> dict:
+    if self._data is None:
+        self._data = await workflow.execute_activity(fetch_data, ...)
```
cretz: Note, we'd fail if you tried to do this anyways

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+        return self._queue[0] if self._queue else None
+
+    @workflow.update
+    def dequeue(self) -> str | None:
+        if self._queue:
+            return self._queue.pop(0)
+        return None
+```
+
+### Blocking in Queries
```
cretz: Same for update validators

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+@activity.defn
+async def process_file(path: str) -> str:
+    async with aiofiles.open(path) as f:
+        return await f.read()
+```
+
+### Missing Executor for Sync Activities
+
+```python
+# BAD - Sync activity without executor blocks worker
```
cretz: > blocks worker

No, this is not true. We would raise an exception as soon as you tried to instantiate a worker w/ a sync activity but no executor

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+    client,
+    task_queue="my-queue",
+    activities=[slow_computation],
+    activity_executor=ThreadPoolExecutor(max_workers=10),
+)
+```
+
+## Error Handling
+
+### Swallowing Errors
```
cretz: Isn't this Python general practice? Worth using skills tokens for this?

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+    except ConnectionError:
+        raise ApplicationError("Connection failed", non_retryable=True)
+
+# GOOD - Only permanent failures are non-retryable
+@activity.defn
+async def call_api():
+    try:
+        return await http_client.get(url)
+    except ConnectionError:
+        raise  # Let Temporal retry
```
cretz: Why even catch it at all, just move comment above the return, or say "Let Temporal retry non-invalid-creds errors"

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+    retry_policy=RetryPolicy(maximum_attempts=1),
+)
+
+# GOOD - Resilient to transient failures
+result = await workflow.execute_activity(
+    flaky_api_call,
+    schedule_to_close_timeout=timedelta(minutes=10),
+    retry_policy=RetryPolicy(
+        initial_interval=timedelta(seconds=1),
+        maximum_interval=timedelta(minutes=1),
```
cretz: Maximum interval is aggressive too (will also stop commenting on not using values that are already the default)

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Heartbeating
+
+### Forgetting to Heartbeat Long Activities
+
+```python
+# BAD - No heartbeat, can't detect stuck activities
+@activity.defn
+async def process_large_file(path: str):
+    for chunk in read_chunks(path):
```
cretz:
```suggestion
    async for chunk in read_chunks(path):
```

(same for next one or any that look like they're blocking)

references/python/gotchas.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Testing
+
+### Not Testing Failures
+
+```python
+# Test failure scenarios
+@pytest.mark.asyncio
+async def test_activity_failure_handling():
+    async with await WorkflowEnvironment.start_time_skipping() as env:
```
cretz: Mentioned before, so won't do it elsewhere, but we need to encourage `start_local` unless `start_time_skipping` is _really_ needed. We also need to encourage reuse of the `start_local` environment (something that can't be done with start time skipping)

references/python/observability.md
```diff
@@ -0,0 +1,191 @@
...
+# Configure a custom handler
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter(
+    "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+))
+
+# Apply to Temporal's logger
+temporal_logger = logging.getLogger("temporalio")
+temporal_logger.addHandler(handler)
+temporal_logger.setLevel(logging.INFO)
```
cretz: This doesn't make sense. You'd usually configure the global logging module or root logger.

references/python/observability.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Metrics
+
+### Enabling SDK Metrics
+
+```python
+from temporalio.client import Client
+from temporalio.runtime import Runtime, TelemetryConfig, PrometheusConfig
+
+# Configure Prometheus metrics endpoint
```
cretz: Can clarify this should be done before any code runs. We might want to encourage `Runtime.set_default` with the runtime instead of trying to have everyone remember to set the `runtime` kwarg everywhere

references/python/observability.md
```diff
@@ -0,0 +1,191 @@
...
+### Key SDK Metrics
+
+- `temporal_request` - Client requests to server
+- `temporal_workflow_task_execution_latency` - Workflow task processing time
+- `temporal_activity_execution_latency` - Activity execution time
+- `temporal_workflow_task_replay_latency` - Replay duration
+
+## Tracing
+
+### OpenTelemetry Integration
```
cretz: This is actively being changed FWIW

references/python/observability.md
```diff
@@ -0,0 +1,191 @@
...
+    interceptors=[tracing_interceptor],
+)
+```
+
+## Search Attributes (Visibility)
+
+### Setting Search Attributes at Start
+
+```python
+from temporalio.common import SearchAttributes, SearchAttributeKey
```
cretz: Mentioned before, but `SearchAttributes` is not what is used with `SearchAttributeKey`, that is `TypedSearchAttributes`. The `search_attributes=` kwarg is deprecated.

references/python/observability.md
```diff
@@ -0,0 +1,191 @@
...
+### Upserting Search Attributes from Workflow
+
+```python
+@workflow.defn
+class OrderWorkflow:
+    @workflow.run
+    async def run(self, order: Order) -> str:
+        # Update status as workflow progresses
+        workflow.upsert_search_attributes([
+            (ORDER_STATUS, "processing"),
```
cretz:
```suggestion
            ORDER_STATUS.value_set("processing"),
```

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+### WHEN:
+- **Order approval workflows** - Wait for human approval before proceeding
+- **Live configuration updates** - Change workflow behavior without restarting
+- **Fire-and-forget communication** - Notify workflow of external events
+- **Workflow coordination** - Allow workflows to communicate with each other
+
+**Signals vs Queries vs Updates:**
+- Signals: Fire-and-forget, no response, can modify state
+- Queries: Read-only, returns data, cannot modify state
+- Updates: Synchronous, returns response, can modify state
```
cretz: I don't know if "synchronous" is the right word here. You most definitely can start an update somewhere and wait for result somewhere else. And an update can most definitely take weeks/months to complete.

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+
+### WHY: Break complex workflows into smaller, manageable units with independent failure domains
+### WHEN:
+- **Failure domain isolation** - Child failures don't automatically fail parent
+- **Different retry policies** - Each child can have its own retry configuration
+- **Reusability** - Share workflow logic across multiple parent workflows
+- **Independent scaling** - Child workflows can run on different task queues
+- **History size management** - Each child has its own event history
+
+**Use activities instead when:** Operation is short-lived, doesn't need its own failure domain, or doesn't need independent retry policies.
```
cretz: Need another "Use normal methods when: ...". Child workflows aren't needed unless their features are needed.

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+        workflow.execute_activity(
+            process_item, item,
+            schedule_to_close_timeout=timedelta(minutes=5)
+        )
+        for item in items
+    ]
+    return await asyncio.gather(*tasks)
+```
+
+### Deterministic Alternatives to asyncio
```
cretz, line 166: May be worth a quick clarification that this is because these two calls in CPython use `set` which is non-deterministic

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+
+```python
+# workflow.wait() - like asyncio.wait()
+done, pending = await workflow.wait(
+    futures,
+    return_when=workflow.WaitConditionResult.FIRST_COMPLETED
+)
+
+# workflow.as_completed() - like asyncio.as_completed()
+for future in workflow.as_completed(futures):
```
cretz:
```suggestion
async for future in workflow.as_completed(futures):
```

I think we need something confirming our snippets work. This has been a problem for docs too and has encouraged them to move to a more "snip sync" type of approach

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+@workflow.run
+async def run(self, state: WorkflowState) -> str:
+    while True:
+        state = await process_batch(state)
+
+        if state.is_complete:
+            return "done"
+
+        # Continue with fresh history before hitting limits
+        if workflow.info().get_current_history_length() > 10000:
```
cretz:
```suggestion
        if workflow.info().is_continue_as_new_suggested():
```

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+
+## Continue-as-New
+
+### WHY: Prevent unbounded event history growth in long-running or infinite workflows
+### WHEN:
+- **Event history approaching 10,000+ events** - Temporal recommends continue-as-new before hitting limits
+- **Infinite/long-running workflows** - Polling, subscription, or daemon-style workflows
+- **Memory optimization** - Reset workflow state to reduce memory footprint
+
+**Recommendation:** Check history length periodically and continue-as-new around 10,000 events.
```
cretz: The recommendation is usually to use `is_continue_as_new_suggested`, which we set after so many (or so large)

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+@workflow.run
+async def run(self, order: Order) -> str:
+    compensations: list[Callable[[], Awaitable[None]]] = []
+
+    try:
+        await workflow.execute_activity(
+            reserve_inventory, order,
+            schedule_to_close_timeout=timedelta(minutes=5)
+        )
+        compensations.append(lambda: workflow.execute_activity(
```
cretz: Depends on use case, but often compensations need to be added before the activity, because what if the activity code succeeds but the failure is in reporting to server or racy timeout after complete something? https://temporal.io/blog/compensating-actions-part-of-a-complete-breakfast-with-sagas#a-quick-note-on-ordering-of-compensations covers this a bit. At the least, AI should be told about why you might want to.

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+        workflow.logger.error(f"Order failed: {e}, running compensations")
+        for compensate in reversed(compensations):
+            try:
+                await compensate()
+            except Exception as comp_err:
+                workflow.logger.error(f"Compensation failed: {comp_err}")
+        raise
+```
+
+## Cancellation Handling
```
cretz: Needs to clarify somewhere this leverages traditional asyncio cancellation. There are some quirks about asyncio cancellation being an event instead of a state like every other SDK, but not sure if we want to get into those details.

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+- **Checkpointing** - Save progress periodically
+
+```python
+@activity.defn
+async def process_large_file(file_path: str) -> str:
+    # Get heartbeat details from previous attempt (if any)
+    heartbeat_details = activity.info().heartbeat_details
+    start_line = heartbeat_details[0] if heartbeat_details else 0
+
+    with open(file_path) as f:
```
cretz: There is a skill earlier about not doing this exact type thing in an `async def`

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+
+## Local Activities
+
+### WHY: Reduce latency for short, lightweight operations by skipping the task queue
+### WHEN:
+- **Short operations** - Activities completing in milliseconds/seconds
+- **High-frequency calls** - When task queue overhead is significant
+- **Low-latency requirements** - When you can't afford task queue round-trip
+
+**Note:** Local activities are experimental in Python SDK.
```
cretz: This is not true, and if it's documented as such somewhere, we should fix it

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+    # Wait for 1 hour
+    await asyncio.sleep(3600)
+
+    # Or with workflow-specific API
+    await workflow.sleep(timedelta(hours=1))
+
+    return "Timer fired"
+```
+
+## Local Activities
```
cretz, line 376: We need to explain why these are discouraged most of the time

references/python/patterns.md
```diff
@@ -0,0 +1,453 @@
...
+        quick_lookup,
+        "key",
+        schedule_to_close_timeout=timedelta(seconds=5),
+    )
+    return result
+```
+
+## Using Pydantic Models
+
+```python
```
cretz: Separate snippets from workflow and non-workflow so not tempted to put these in the same file (just separate code fences is fine)

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+from temporalio.worker import Worker
+
+from activities.greet import greet
+from workflows.greeting import GreetingWorkflow
+
+async def main():
+    client = await Client.connect("localhost:7233")
+    async with Worker(client, task_queue="greeting-queue",
+                      workflows=[GreetingWorkflow], activities=[greet]):
+        result = await client.execute_workflow(
```
cretz: So this is actually an uncommon approach that makes sense only in single-file samples and tests. The starter of the workflow is usually wholly divorced from the worker. Would recommend separating worker and starter.

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+- Can be sync or async functions
+- **Default to sync activities** - safer and easier to debug
+- Sync activities need `activity_executor` (ThreadPoolExecutor)
+- Async activities require async-safe libraries throughout (e.g., `aiohttp` not `requests`)
+
+See `sync-vs-async.md` for detailed guidance on choosing between sync and async.
+
+### Worker Setup
+- Connect client, create Worker with workflows and activities
+- Use `async with Worker(...)` context manager
```
cretz: Not necessarily, `await my_worker.run()` is plenty normal

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+The Python SDK's sandbox provides automatic protection against many common non-deterministic operations, but understanding replay helps you write correct code.
+
+See `determinism.md` for detailed rules and safe alternatives.
+
+## Determinism Rules Summary
+
+**Safe alternatives:**
+- `workflow.now()` instead of `datetime.now()`
+- `workflow.random()` instead of `random`
+- `asyncio.sleep()` works (sandbox handles it)
```
cretz: There actually is a `workflow.sleep` and we often encourage it because you can set a summary for your timer

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+- `workflow.uuid4()` for UUIDs
+- `workflow.wait()` instead of `asyncio.wait()` (deterministic ordering)
+- `workflow.as_completed()` instead of `asyncio.as_completed()` (deterministic ordering)
+
+**Pass-through for libraries:**
+```python
+with workflow.unsafe.imports_passed_through():
+    import pydantic
```
cretz: This is a poor example, `pydantic` is already passed through by the system

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+## Replay-Aware Logging
```
cretz: Lot of duplicate stuff across these skills

references/python/python.md
```diff
@@ -0,0 +1,176 @@
...
+class MyWorkflow:
+    @workflow.run
+    async def run(self) -> str:
+        workflow.logger.info("Starting workflow")  # Only logs on initial execution
+        result = await workflow.execute_activity(...)
+        workflow.logger.info(f"Activity result: {result}")
+        return result
+```
+
+Use `activity.logger` in activities for context-aware logging with automatic correlation:
```
cretz: Doesn't make sense in a section entitled "Replay-Aware Logging"

references/python/sandbox.md
```diff
@@ -0,0 +1,180 @@
...
+result = some_library.cached_operation()  # Cache changes between replays
+```
+
+**Fix**: Move to activity or use pass-through with caution.
+
+### Slow Worker Startup
+
+Large workflow files slow down worker initialization because they're reloaded frequently.
+
+**Fix**: Keep workflow files minimal, move logic to activities.
```
cretz: I don't think large workflow files slow down anything hardly at all, but workflows that import lots of things non-passthrough do. We should not necessarily encourage people to move what could be durable logic to activities (sometimes we should, but it has nothing to do with file size).

references/python/sync-vs-async.md
```diff
@@ -0,0 +1,243 @@
...
+- Worker cannot communicate with Temporal Server
+- Workflow progress blocks across the worker
+- Potential deadlocks and unpredictable behavior
+- Difficult-to-diagnose bugs
+
+## How the SDK Handles Each Type
+
+### Synchronous Activities
+
+- Run in the `activity_executor` (thread pool by default)
```
cretz: There is no default

references/python/sync-vs-async.md
```diff
@@ -0,0 +1,243 @@
...
+3. If bugs disappear, the original async activity had blocking calls
+
+## Threading Considerations
+
+### Multi-Core Usage
+
+For CPU-bound work or true parallelism:
+
+- Use multiple worker processes
+- Or use `ProcessPoolExecutor` for synchronous activities
```
cretz: We should discourage this or at least explain some downsides (e.g. cancellation delivery)

references/python/testing.md
```diff
@@ -0,0 +1,120 @@
...
+                "input",
+                id="test-workflow-id",
+                task_queue="test-queue",
+            )
+            assert result == "expected"
+```
+
+## Local Test Environment
+
+For tests that don't need time-skipping:
```
cretz: Specifically, should say in the previous section that Time-Skipping Test Environment should only be when you _must_ have time-skipping

references/python/testing.md
```diff
@@ -0,0 +1,120 @@
...
+
+```python
+from temporalio.testing import ActivityEnvironment
+
+@pytest.mark.asyncio
+async def test_activity():
+    env = ActivityEnvironment()
+
+    # Optionally customize activity info
+    # env.info = ActivityInfo(...)
```
cretz: We have a newer suggestion which is using `dataclasses.replace` with the `ActivityEnvironment.default info()`

references/python/testing.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Mocking Activities
+
+```python
+async def mock_activity(input: str) -> str:
+    return "mocked result"
+
+@pytest.mark.asyncio
+async def test_with_mock():
+    async with await WorkflowEnvironment.start_time_skipping() as env:
```
cretz: Mentioned before, but we should stop using time skipping unless we really need it, it's a poor test env (can't be reused, limited functionality, etc)

cretz:
Note, I did not review TypeScript, only core and Python.

references/core/gotchas.md
donald-pinckney: Lifted up to a top-level issue to keep track of later. Not changing this currently.

references/core/determinism.md
donald-pinckney: Added Go. Made an issue for the other languages for the future.

references/core/interactive-workflows.md
donald-pinckney: I've cleaned it up to be reasonable and well-defined. I'm keeping the content here for now though. Testing interactive workflows is not the same as testing other workflows, since you have to know to send updates / signals to it during the execution. Some of this is duplicative with general workflow testing though, may move it in future.

references/core/patterns.md
donald-pinckney: Yup. Fixed.

references/core/tool-reference.md
donald-pinckney: Agreed. Cutting much / all of this.

donald-pinckney: Cut all the shell scripts, made an issue to re-introduce some of that functionality (via coaching it to use CLI).

references/core/tool-reference.md
donald-pinckney: Reworded and clarified it.

references/python/advanced-features.md
donald-pinckney: Agreed. I'm going to cut this for this release, and add proper plugin info in future.

references/core/versioning.md
donald-pinckney: Put into a new issue to keep track of.

references/python/python.md
donald-pinckney: Unsure if thats true, the docs currently have the same example: https://docs.temporal.io/develop/python/python-sdk-sandbox#passthrough-modules

Will make an issue to verify which is true, and if you're right, update docs too.

references/python/patterns.md
donald-pinckney: Yup, agreed. I'm also concerned about references in skill files to other skill files: those should be validated as well that they don't "404". Will think about options for these.

references/python/patterns.md
donald-pinckney: Yup, it was documented as such on the python sdk readme. Fix: https://github.com/temporalio/sdk-python/pull/1334

references/python/observability.md
donald-pinckney: FYI, it sure doesn't look like the `search_attributes` kwarg on start_workflow is deprecated. Its specifically recommended in the docs (https://docs.temporal.io/develop/python/observability#custom-search-attributes).

Have fixed the `TypedSearchAttributes` stuff though.

references/python/observability.md
donald-pinckney: Yeah, i cut it for now.

references/python/testing.md
donald-pinckney: @MasonEgger this one is interesting, as courses and learn.temporal.io seem to push `start_time_skipping` rather than `start_local`:
- https://learn.temporal.io/getting_started/python/hello_world_in_python/#-add-a-unit-test
- https://github.com/temporalio/edu-102-python-content/blob/main/testing-application-code/testing-workflows.md

As a result the Kappa agent for Temporal docs suggests using time skipping over non-time skipping. Do you have any thoughts to add here?

MasonEgger: This was what we recommended in courses because our content did have timers in it. I default to Chad on this, but I'd like this to be noted in this skill so the Skill knows when to use it. I was unaware of the poor test env scenarios.

donald-pinckney: Independent of the skill, its also an issue for Kapa. If I prompt Kapa:

> How do i write tests for workflows in python?

It will strongly push time-skipping as the default to use. I think Kapa is picking up the context of those test examples from the courses, but not seeing that those are only like that because of timers elsewhere in the courses.

donald-pinckney: Part of this is that the normal Python docs don't mention `WorkflowEnvironment` at all (except under time skipping section!). If those were updated to explain the standard setup for how to use `WorkflowEnvironment` when writing tests, perhaps Kappa could pull from those rather than courses...

https://docs.temporal.io/develop/python/testing-suite

Issue #23, opened 2026-02-19
Add otel


donald-pinckney:
Originally had otel, but cut it in commit f455d39af8e856422b72330d70d7d8daf23763d4 in order to get to something shippable that is correct faster.

Currently otel is not mentioned for any of the languages, but it should be described in observability.md.

_Originally posted by @donald-pinckney in https://github.com/temporalio/skill-temporal-developer/pull/4#discussion_r2829018941_

PR #31, opened 2026-02-23
Add TypeScript

donald-pinckney:
## TypeScript support for `temporal-developer` skill!

This PR adds TypeScript support into the skill. Its structure is parallel and similar to Python's.

Several automated agentic review passes were done on this content to check for content alignment with `python` and `core`, and to compare against docs for factual accuracy.

### DIFF
--- a/references/typescript/advanced-features.md
+++ b/references/typescript/advanced-features.md
@@ -51,12 +51,12 @@ export async function doSomethingAsync(): Promise<string> {
 
 **External completion (from another process, machine, etc.):**
 ```typescript
-import { AsyncCompletionClient } from '@temporalio/client';
+import { Client } from '@temporalio/client';
 
 async function doSomeWork(taskToken: Uint8Array): Promise<void> {
-  const client = new AsyncCompletionClient();
+  const client = new Client();
   // does some work...
-  await client.complete(taskToken, "Job's done!");
+  await client.activity.complete(taskToken, "Job's done!");
 }
 ```
 


references/typescript/testing.md
```diff
@@ -0,0 +1,222 @@
+# TypeScript SDK Testing
+
+## Overview
+
+The TypeScript SDK provides `TestWorkflowEnvironment` for testing workflows with time-skipping and activity mocking support. Use `createTimeSkipping()` for automatic time advancement when testing workflows with timers, or `createLocal()` for a full local server without time-skipping.
+
+**Note:** Prefer to use `createLocal()` for full-featured support. Only use `createTimeSkipping()` if you genuinely need time skipping for testing your workflow.
```
donald-pinckney, line 7: Is this correct? Like it is in Python, is it also preferable to use `createLocal` rather than `createTimeSkipping` in TypeScript? (unless you need time skipping).

references/typescript/advanced-features.md
```diff
@@ -0,0 +1,150 @@
...
+  throw new CompleteAsyncError();
+}
+```
+
+**External completion (from another process, machine, etc.):**
+```typescript
+import { AsyncCompletionClient } from '@temporalio/client';
+
+async function doSomeWork(taskToken: Uint8Array): Promise<void> {
+  const client = new AsyncCompletionClient();
```
donald-pinckney: Is `new AsyncCompletionClient()` or `new Client().activity` preferred here? Official sources are contradictory actually. 

- [API Docs](https://typescript.temporal.io/api/classes/client.AsyncCompletionClient) say "Typically AsyncCompletionClient should not be instantiated directly"
- The [docs / samples](https://github.com/temporalio/samples-typescript/blob/main/activities-examples/src/activities/async-completion.ts) only show directly instantiating `AsyncCompletionClient`

SKILL.md
```diff
@@ -108,7 +109,7 @@ Once you've downloaded the file, extract the downloaded archive and add the temp
 - **`references/core/interactive-workflows.md`** - Testing signals, updates, queries
 - **`references/core/dev-management.md`** - Dev cycle & management of server and workers
 - **`references/core/ai-patterns.md`** - AI/LLM pattern concepts
-    + Langauge-specific info at `references/{your_language}/determinism.md`
+    + Langauge-specific info at `references/{your_language}/determinism.md`, if available. Currently Python only.
```
mjameswh:
```suggestion
    + Language-specific info at `references/{your_language}/determinism.md`, if available. Currently Python only.
```

references/typescript/typescript.md
```diff
@@ -0,0 +1,168 @@
+# Temporal TypeScript SDK Reference
+
+## Overview
+
+The Temporal TypeScript SDK provides a modern async/await approach to building durable workflows. Workflows run in an isolated runtime with bundling and automatic replacements for determinism protection.
```
chris-olszewski:
```suggestion
The Temporal TypeScript SDK provides a modern Promise based approach to building durable workflows. Workflows are bundled and run in an isolated runtime with automatic replacements for determinism protection.
```

references/typescript/typescript.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Understanding Replay
+
+Temporal workflows are durable through history replay. For details on how this works, see `references/core/determinism.md`.
+
+## Quick Start
+
+**Add Dependencies:** Install the Temporal SDK packages:
+```bash
+npm install @temporalio/client @temporalio/worker @temporalio/workflow @temporalio/activity
```
chris-olszewski, line 17: Not sure the best way to encode this, but users might not always use `npm`, there is a lot of competition in the JS package manager space. e.g. if using `pnpm` the command should be `pnpm add`

references/typescript/typescript.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+**activities.ts** - Activity definitions (separate file for bundling performance):
```
chris-olszewski: To my knowledge activities are defined in a separate file purely to isolate activity dependencies from workflow ones
```suggestion
**activities.ts** - Activity definitions (separate file to distinguish workflow vs activity code):
```

references/typescript/typescript.md
```diff
@@ -0,0 +1,168 @@
...
+    activities,
+    taskQueue: 'greeting-queue',
+  });
+  await worker.run();
+}
+
+run().catch(console.error);
+```
+
+**Start the dev server:** Start `temporal server start-dev` in the background.
```
chris-olszewski, line 60: Not sure if it matter, but users might already have this running which would cause this command to fail.

references/typescript/typescript.md
```diff
@@ -0,0 +1,168 @@
...
+
+See `references/typescript/determinism.md` for detailed rules.
+
+## Common Pitfalls
+
+1. **Importing activities without `type`** - Use `import type * as activities`
+2. **Version mismatch** - All @temporalio packages must match
+3. **Direct I/O in workflows** - Use activities for external calls
+4. **Missing `proxyActivities`** - Required to call activities from workflows
+5. **Forgetting to bundle workflows** - Worker needs workflowsPath
```
chris-olszewski:
```suggestion
5. **Forgetting to bundle workflows** - Worker needs workflowsPath or workflowBundle
```

references/typescript/gotchas.md
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// GOOD - Use workflow sleep
+import { sleep } from '@temporalio/workflow';
+
+export async function delayedWorkflow(): Promise<void> {
+  await sleep('1 minute'); // Durable, survives restarts
+  await activities.doWork();
+}
+```
```
chris-olszewski: Another gotcha that might make sense to call out would be using `workflowPath` instead of `workflowBundle` for production. `workflowPath` should only be used for testing and development.

references/typescript/observability.md
```diff
@@ -0,0 +1,126 @@
...
+  },
+});
+```
+
+## Best Practices
+
+1. Use `log` from `@temporalio/workflow` in workflows - workflows run in a sandbox and need replay-aware logging
+2. Include correlation IDs (orderId, customerId) in log messages
+3. Configure Winston or similar for production log aggregation
+4. Enable OpenTelemetry for distributed tracing across services
```
chris-olszewski: There should be a section on using `@temporalio/interceptors-opentelemetry`: https://github.com/temporalio/samples-typescript/tree/main/interceptors-opentelemetry

references/typescript/determinism.md
```diff
@@ -0,0 +1,73 @@
...
+  const childWorkflowId = uuid4();
+  await executeChild(childWorkflow, {
+    workflowId: childWorkflowId,
+    args: [input],
+  });
+}
+```
+
+**When to use:**
+- Generating unique IDs for child workflows
```
chris-olszewski: Unsure if important, but `workflowId` is an optional param that uses `uuid4` if not provided

references/typescript/determinism.md
```diff
@@ -0,0 +1,73 @@
...
+- Creating idempotency keys
+- Any situation requiring unique identifiers in workflow code
+
+## Forbidden Operations
+
+```typescript
+// DO NOT do these in workflows:
+import fs from 'fs';           // Node.js modules
+fetch('https://...');          // Network I/O
+console.log();                 // Side effects (use workflow.log)
```
chris-olszewski: I think this is fine to use in workflow code, we patch out all `console.log|warn|error` functions to include workflow id

references/typescript/typescript.md
donald-pinckney: Yes, good call.

references/typescript/gotchas.md
donald-pinckney: Oh very good point. I hadn't known that. Will probably update that throughout.

references/typescript/observability.md
donald-pinckney: Ah yes, for the present I'm not including otel information in the skill, but will add it in a follow-up (https://github.com/temporalio/skill-temporal-developer/issues/23).

references/typescript/typescript.md
donald-pinckney: I think it is ok, the files in core describe this carefully enough, so i think its ok to be a bit imprecise here.

references/typescript/typescript.md
donald-pinckney: Added a small caveat. That should be sufficient.

references/typescript/determinism.md
donald-pinckney: I think its better to have claude keep workflow IDs explicit in their creation. Regardless, I've cut this section, IMO its too many tokens for such a simple thing.

references/typescript/determinism-protection.md
```diff
@@ -0,0 +1,56 @@
...
+**Note**: Modules with the `node:` prefix (e.g., `node:fs`) require additional webpack configuration to ignore. You may need to configure the bundler's `externals` or use webpack `resolve.alias` to handle these imports.
+
+Use this with *extreme caution*.
+
+
+## Function Replacement
+
+Functions like `Math.random()`, `Date`, and `setTimeout()` are replaced by deterministic versions.
+
+Date-related functions will *deterministically* return the date at the *start of the workflow*, and will only progress in time when a semantic time operation occurs in Temporal, like a durable sleep. For example:
```
mjameswh: I think the phrasing here is a bit hard to understand correctly. I'd suggest something more along of "the timestamp at which the current workflow task was initially executed; i.e. that timestamp will remain the same when the workflow task is replayed".

mjameswh: I think the phrasing here is a bit hard to understand correctly. I'd suggest something more along of "the timestamp at which the current workflow task was initially executed; i.e. that timestamp will remain the same when the workflow task is replayed".

references/typescript/determinism-protection.md
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// this prints timestamps increasing roughly 1s each iteration
+for (let x = 0; x < 10; ++x) {
+  await sleep('1 second');
+  console.log(Date.now());
+}
+```
+
+This means that if you want a workflow to truly be able to check current real-world physical time, you should retrieve the time in an activity. You should consider which is semantically appropriate for your situation.
```
mjameswh: That comment makes no sense IMHO. If calling an activity to get current time, the activity will get scheduled and executed at a later time. Then, the workflow's activity call will, again, resolve at a later time. And in the future, the activity will always resolve with the same timestamp on replay.

There's no case where "retrieve the time in an activity" will provide a better estimatino of "current real-world physical time" than simply calling `Date.now()`.

If one _really_ needs to get "the current real-world physical time" form workflow code, in TS, it is possible to call `workflowInfo().unsafe.now()`. But this is very rarely the correct thing to do, and must never be used in a way that would impact the workflow code execution. One valid use cases for this is some query handlers or debugging log statements.

references/typescript/determinism.md
```diff
@@ -0,0 +1,51 @@
...
+## Why Determinism Matters
+
+Temporal provides durable execution through **History Replay**. When a Worker needs to restore workflow state (after a crash, cache eviction, or to continue after a long timer), it re-executes the workflow code from the beginning, which requires the workflow code to be **deterministic**.
+
+## Temporal's V8 Sandbox
+
+The Temporal TypeScript SDK executes all workflow code in sandbox, which (among other things), replaces common non-deterministic functions with deterministic variants. As an example, consider the code below:
+
+```ts
+export async function badWorkflow(): Promise<string> {
```
mjameswh: Not sure why this workflow is named "bad". It will work perfectly in TS, thanks to the sandbox.

references/typescript/determinism.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Testing Replay Compatibility
+
+Use `Worker.runReplayHistory()` to verify your code changes are compatible with existing histories. See the Workflow Replay Testing section of `references/typescript/testing.md`.
+
+## Best Practices
+
+1. Use type-only imports for activities in workflow files
+2. Match all @temporalio package versions
+3. Use `sleep()` from workflow package, not `setTimeout` directly
```
mjameswh: It's true that `sleep()` is "preferred", but I'm concerned that the current phrasing might make agents flag usage of `setTimeout` as a potential source of bug. There's really nothing wrong with `setTimeout` and `clearTimeout`, except that it's less obvious for users how those interract with cancellation scopes.

Without entering to details, I'd suggest softening language a bit on this item.

mjameswh (commented):
General: `peristed` (in determinism.md)

references/typescript/determinism-protection.md
donald-pinckney: HHmm. I agree it doesn't currently make much sense. And probably it would be best to just remove this sentence.

That being said, there is something semantically distinct with putting Date.now() in the workflow vs. activity. If you do:

```ts
const t0 = Date.now() // or, const t0 = await getDateActivity()

// ... some code C that takes 10 seconds to run ...

const t1 = Date.now() // or, const t1 = await getDateActivity()
```

If Date.now() is in the workflow, then t0 == t1, as long as C doesn't do any Temporal calls. But if you put the Date.now() calls into activities, then t1 > t0. But this really only applies if you are blocking the workflow with compute, which is bad anyways.

references/typescript/advanced-features.md
mjameswh: The recommended pattern is to create `Client` first, then access subclients as properties on that client.

```
const client = new Client({ ... });

client.activity.xxxx();
client.workflow.xxx();
...
```

This has been the recommended approach for ~3 years. Code that still instantiate a subclient directly is outdated.

references/typescript/determinism-protection.md
donald-pinckney: Removed this comment.

references/typescript/data-handling.md
```diff
@@ -0,0 +1,280 @@
...
+export async function orderWorkflow(): Promise<void> {
+  const info = workflowInfo();
+  const customerName = info.memo?.customerName;
+  // ...
+}
+```
+
+## Protobuf Support
+
+Using Protocol Buffers for type-safe serialization.
```
mjameswh: Please mention somewhere that JSON serialization is preferred in TS apps. For context, it is significantly simpler to implement, but also, serializating to protobuf significantly increase the per-workflow execution memory footprint, which reduces the number of workflow they may cache and the time it takes to load a workflow sandbox.

references/typescript/determinism.md
donald-pinckney: Yup, it is not bad at all. Fixed.

references/typescript/data-handling.md
```diff
@@ -0,0 +1,280 @@
...
+const dataConverter: DataConverter = {
+  payloadConverter: new DefaultPayloadConverterWithProtobufs({
+    protobufRoot: myProtobufRoot,
+  }),
+};
+```
+
+## Large Payloads
+
+For large data, consider:
```
mjameswh: We need to clarify somewhere what "large" means in the context of a workflow. There are several limits that applies:
- max 2MB per payload
- max 4MB per grpc message
- max 50MB for the complete workflow history (but in practice, should aim to stay below 10MB).

references/typescript/data-handling.md
```diff
@@ -0,0 +1,280 @@
...
+
+```typescript
+// Example: Reference pattern for large data
+import { proxyActivities } from '@temporalio/workflow';
+
+const { uploadToStorage, downloadFromStorage } = proxyActivities<Activities>({
+  startToCloseTimeout: '5 minutes',
+});
+
+export async function processLargeDataWorkflow(dataRef: string): Promise<void> {
```
mjameswh: This is an antipattern if the supposition is that `data` is potentially large. This is bringing the large data into workflow history, twice (once as return value of the `downloadFromStorage` activity, and then as arg to the `processData` activity).

We'll need a better example for this.

Also, I'd recommend presenting this in a language agnostic doc. That's part of general Temporal usage ptterns.

references/typescript/data-handling.md
```diff
@@ -0,0 +1,280 @@
+# TypeScript SDK Data Handling
+
+## Overview
```
mjameswh, line 3: The term Payload is never properly introduced, and I see no explanation on what are payloadconverters vs payloadcodecs. This distinction is particularly important in sandboxed languages (TS and Python).

references/typescript/error-handling.md
```diff
@@ -0,0 +1,118 @@
...
+
+const { riskyActivity } = proxyActivities<typeof activities>({
+  startToCloseTimeout: '5 minutes',
+});
+
+export async function workflowWithErrorHandling(): Promise<string> {
+  try {
+    return await riskyActivity();
+  } catch (err) {
+    if (err instanceof ApplicationFailure) {
```
mjameswh, line 51: IIRC, that `err` would be an `ActivityFailure`, with `cause` being the `ApplicationFailure`.

references/typescript/error-handling.md
```diff
@@ -0,0 +1,118 @@
...
+    initialInterval: '1s',
+    backoffCoefficient: 2,
+    maximumInterval: '1m',
+    maximumAttempts: 5,
+    nonRetryableErrorTypes: ['ValidationError', 'PaymentError'],
+  },
+});
+```
+
+**Note:** Only set retry options if you have a domain-specific reason to. The defaults are suitable for most use cases.
```
mjameswh, line 74: But should always set either `scheduleToCloseTimeout` or `startToCloseTimeout`. Beware that `startToCloseTimeout` alone could cause retrying forever; combine with `maxAttempts` or `scheduleToStartTimeout` if retying forever is undesirable. Note that it is not uncommon for Temporal workflows to actually _want_ to retry failing activities forever.

references/typescript/error-handling.md
```diff
@@ -0,0 +1,118 @@
...
+    throw ApplicationFailure.create({
+      message: 'Workflow failed due to invalid state',
+      type: 'InvalidStateError',
+    });
+  }
+  return 'success';
+}
+```
+
+**Warning:** Do NOT use `nonRetryable: true` for workflow failures in most cases. Unlike activities, workflow retries are controlled by the caller, not retry policies. Use `nonRetryable` only for errors that are truly unrecoverable (e.g., invalid input that will never be valid).
```
mjameswh, line 104: Ehh? Throwing an `ApplicationFailure` from a workflow will _always_ fail the workflow. `ApplicationFailure.nonRetryable` has no impact in when thrown from workflows. To cause wokrflow task failure (so that the wft will get retried), throw a non-`TemporalFailure` error (e.g. `throw new Error(...)`).

Also, it's true that caller has some control on "workflow retry", but this mechanism is non-intuitive and often misunderstood. We generally recommend to not depend on workflow retries.

references/typescript/error-handling.md
```diff
@@ -0,0 +1,118 @@
...
+## Idempotency
+
+For idempotency patterns (using keys, making activities granular), see `core/patterns.md`.
+
+## Best Practices
+
+1. Use specific error types for different failure modes
+2. Set `nonRetryable: true` for permanent failures in activities
+3. Configure `nonRetryableErrorTypes` in retry policy
+4. Always re-throw errors after handling in workflows
```
mjameswh: Why "always"?

references/typescript/typescript.md
```diff
@@ -0,0 +1,169 @@
...
+```typescript
+import { Client } from '@temporalio/client';
+import { greetingWorkflow } from './workflows';
+import { v4 as uuid } from 'uuid';
+
+async function run() {
+  const client = new Client();
+
+  const result = await client.workflow.execute(greetingWorkflow, {
+    workflowId: uuid(),
```
mjameswh, line 74: Using uuids for workflow id is generally perceived as an antipattern. Yes, lots of our samples do that, but still something we're trying to move away. Workflow ids should be business meaningful as much as possible.

references/typescript/typescript.md
```diff
@@ -0,0 +1,169 @@
...
+## Common Pitfalls
+
+1. **Importing activities without `type`** - Use `import type * as activities`
+2. **Version mismatch** - All @temporalio packages must match
+3. **Direct I/O in workflows** - Use activities for external calls
+4. **Missing `proxyActivities`** - Required to call activities from workflows
+5. **Forgetting to bundle workflows** - Worker needs `workflowsPath` or `workflowBundle`
+6. **Using workflowsPath in production** - Use `workflowBundle` for production (see `references/typescript/gotchas.md`)
+7. **Forgetting to heartbeat** - Long-running activities need `heartbeat()` calls
+8. **Logging in workflows** - Prefer `import { log } from '@temporalio/workflow'` for replay-aware logging with workflow context. `console.log` also works (it's patched to include workflow ID). See `references/typescript/observability.md`.
```
mjameswh: `console.log()` works, but isn't forwarded to the runtime's logger. The `log` api should be preferred.

...That is, except when instrumenting workflow code for debugging purpose! `console.log()` is sent directly and immediately sent to the console, whereas `log.xxx()` goes through sinks, which may result in messages to be lost in some cases of workflow errors.

references/typescript/typescript.md
```diff
@@ -0,0 +1,169 @@
...
+- `setTimeout` → deterministic timer
+
+**Safe to use:**
+- `sleep()` from `@temporalio/workflow`
+- `condition()` for waiting
+- Standard JavaScript operations
+
+See `references/typescript/determinism.md` for detailed rules.
+
+## Common Pitfalls
```
mjameswh, line 144: Another common issue: missing `await` on activity calls and other commands.

references/typescript/testing.md
mjameswh: That's arguable. What you write is in line with my personal opinion, because the the CLI dev server is the real server, whereas the time skipping server is a distinct, java-based implementatino of the server, and consequently have a certain number of inconsistent behaviors.

But from end user's perspective, testing with the real server (e.g. `createLocal()`) is often impractical, notably when they have long timers.

references/typescript/testing.md
```diff
@@ -0,0 +1,222 @@
...
+  });
+
+  it('replays workflow history from server', async () => {
+    // Fetch history programmatically using the client
+    const connection = await Connection.connect({ address: 'localhost:7233' });
+    const client = new Client({ connection, namespace: 'default' });
+    const handle = client.workflow.getHandle('my-workflow-id');
+    const history = await handle.fetchHistory();
+
+    await Worker.runReplayHistory(
```
mjameswh, line 172: Here, I'd recommend giving a sample for the batch replayer api rather than the "single execution" replayer. Use a `client.workflow.list({ ... })` above to find workflows that match some request.

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+  setHandler(addItemSignal, (item) => {
+    items.push(item);
+  });
+
+  await condition(() => approved);
+  return `Processed ${items.length} items`;
+}
+```
+
+## Dynamic Signal Handlers
```
mjameswh, line 28: Might want to indicate that use cases for this pattern are rare.

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+    progress = i;
+    await doWork();
+  }
+  status = 'completed';
+}
+```
+
+## Dynamic Query Handlers
+
+For handling queries with names not known at compile time:
```
mjameswh: Rare.

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+    return 'approved';
+  } else {
+    return 'auto-rejected due to timeout';
+  }
+}
+```
+
+## Waiting for All Handlers to Finish
+
+### WHY: Ensure all signal/update handlers complete before workflow exits
```
mjameswh: This is missing some context.

We generally recommend that signals and update handlers be non-async. For example, it is prefeerable not to run activities from a signal/update handler. Otherwise, and unless other precautions are taken, there is a risk that the workflow may complete before the handler completes its execution. However, reach that goal sometime requires non-negligeable workarounds and add complexity.

In those cases, one may use `await condition(allHandlersFinished)` at the end of a workflow (or right before calling continue-as-new) to prevent completion of the workflow until all pending async handlers complete.

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+  // Before exiting, wait for all handlers to finish
+  await condition(allHandlersFinished);
+  return 'done';
+}
+```
+
+## Activity Heartbeat Details
+
+### WHY: Resume activity progress after worker failure
+### WHEN:
```
mjameswh, line 352: Missing the most important case: "Support activity cancellation".

The code below should be updated to demonstrate handling cancellation.

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+    await processLine(lines[i]);
+    // Heartbeat with progress
+    heartbeat(i + 1);
+  }
+
+  return 'completed';
+}
+```
+
+## Timers
```
mjameswh, line 385: This is an advanced case of timers. I'd call this section "Cancellable Timers".

references/typescript/patterns.md
```diff
@@ -0,0 +1,414 @@
...
+
+```typescript
+import { sleep, CancellationScope } from '@temporalio/workflow';
+
+export async function timerWorkflow(): Promise<string> {
+  await sleep('1 hour');
```
mjameswh, line 391: This workflow is `sleep()`'ing twice, for a total of 2 hours. That's very likely undesirable.

references/typescript/gotchas.md
```diff
@@ -0,0 +1,274 @@
...
+    "@temporalio/client": "1.9.0",
+    "@temporalio/worker": "1.8.0",
+    "@temporalio/workflow": "1.9.1"
+  }
+}
+
+// GOOD - All versions match
+{
+  "dependencies": {
+    "@temporalio/client": "1.9.0",
```
mjameswh: Recommendation is to specify version as `~1.15.0` (allow patch level upgrade, recommended for production apps), or `^1.15.0` (for non production apps, allowing for minor version bumps).

references/typescript/gotchas.md
```diff
@@ -0,0 +1,274 @@
...
+
+// GOOD: Mark permanent errors as non-retryable
+throw ApplicationFailure.nonRetryable('User not found');
+```
+
+For detailed guidance on error classification and retry policies, see `error-handling.md`.
+
+## Cancellation
+
+### Not Handling Cancellation
```
mjameswh: The snippets given here is only about cancellation, it is about any form of "cleanup on error" handling in workflow code.

I'm pointing this out becasue there is also another case of frequent mistake which is to not handle cancellation _in activities_, which requires 1) heartbeating the activity and 2) awaiting the activity cancellation promise _or_ using the cancellation abort signal.

This should be mentioned here.

references/typescript/gotchas.md
```diff
@@ -0,0 +1,274 @@
...
+export async function processLargeFile(path: string): Promise<void> {
+  let i = 0;
+  for await (const chunk of readChunks(path)) {
+    heartbeat(`Processing chunk ${i++}`);
+    await processChunk(chunk);
+  }
+}
+```
+
+### Heartbeat Timeout Too Short
```
mjameswh, line 237: Heartbeat timeout should be set as high as acceptable for the given use case. There's a cost to heartbeating (each heartbeat counts as one action on Temporal Cloud; in self hosted, that's one persistent mutable state update sent to the database, for each heartbeat).

references/typescript/gotchas.md
```diff
@@ -0,0 +1,274 @@
...
+  );
+});
+```
+
+## Timers and Sleep
+
+### Using JavaScript setTimeout
+
+```typescript
+// BAD - setTimeout is not durable
```
mjameswh: That isn't true. The SDK mocks `setTimeout` and `clearTimeout`. They result in timers, just like `sleep()`.

The only reason we recommend `sleep()` over `setTimeout()` is that `sleep()`'s interraction with cancellation scopes is a bit more intuitive.

references/typescript/versioning.md
```diff
@@ -0,0 +1,202 @@
+# TypeScript SDK Versioning
+
+For conceptual overview and guidance on choosing an approach, see `references/core/versioning.md`.
+
+## Patching API
```
mjameswh, line 5: One key difference in how `patched()` works in TS vs other core based SDKs:
- In Python, .Net and Ruby, `patched()` return value is always memoized on the first call; that means it is not possible to use `patched()` in loops. A common work around is to append a sequence number to patch id which is increased every time the patch is checked.
- In TS, `patched()` is not memoized on negative return. That means it possible to use `patched()` in a loop. However, that also means that user may sometime need to manually memoize the return of `patched()` if a change implies coordinated behavioral changes at different point in the workflow code.

This is a frequent source of confusion. See https://github.com/temporalio/features/issues/591.

references/typescript/versioning.md
```diff
@@ -0,0 +1,202 @@
...
+  }
+}
+```
+
+**How it works:**
+- If the Workflow is running for the first time, `patched()` returns `true` and inserts a marker into the Event History
+- During replay, if the history contains a marker with the same `patchId`, `patched()` returns `true`
+- During replay, if no matching marker exists, `patched()` returns `false`
+
+### Three-Step Patching Process
```
mjameswh, line 39: Should note somewhere that failing to correctly apply this three steps patching process will result in non-deterministic errors.

references/typescript/versioning.md
```diff
@@ -0,0 +1,202 @@
+# TypeScript SDK Versioning
```
mjameswh, line 1: We should add more details regarding what changes requires versioning. Many workflow code changes will not cause NDEs (i.e. changing retry policies on activities, changing a timer's duration, changing args to activities or child workflows, etc). Systematically adding patches in cases where it is not needed adds complexity and can quickly make the workflow code unmanageable.

references/typescript/observability.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Activity Logging
+
+```typescript
+import { log } from '@temporalio/activity';
+
+export async function processPayment(orderId: string): Promise<string> {
+  log.info('Processing payment', { orderId });
+
+  // Activity logs don't need replay suppression
```
mjameswh: > // Activity logs don't need replay suppression

I don't think that's useful to mention.

The main reasons to use activity's `log` api are:
- it adds contextual metadata (activity id, type, namespace, etc) on emitted message
- it funnels emitted log messages to the runtime's logger, making it easier to consistently collect all log messages.

references/typescript/observability.md
```diff
@@ -0,0 +1,110 @@
...
+  },
+});
+```
+
+## Best Practices
+
+1. Prefer `log` from `@temporalio/workflow` in workflows for replay-aware logging with workflow context. `console.log` also works (it's patched to include workflow ID)
+2. Include correlation IDs (orderId, customerId) in log messages
+3. Configure Winston or similar for production log aggregation
+4. Monitor Prometheus metrics for worker health
```
mjameswh, line 108: or OpenTelemetry

references/typescript/data-handling.md
donald-pinckney: Added it in core

references/typescript/data-handling.md
donald-pinckney: I've re-worked this. Fixed the pattern, and lifted it all up to core.

donald-pinckney: Also applied to Python

references/typescript/typescript.md
donald-pinckney: Hhmm. But for writing workflow tests, would we still consider UUIDs as a fine/good pattern? In those cases you may not have a business-logic ID. Or would we suggest at least something like `order-id-{uuid()}`.

references/typescript/typescript.md
donald-pinckney: Improved it, with guidance: observability -> use log. debugging -> console.log is ok.

references/typescript/typescript.md
donald-pinckney: Added it (and to Python)

references/typescript/observability.md
donald-pinckney: Deferred for the moment (https://github.com/temporalio/skill-temporal-developer/issues/23).

references/typescript/versioning.md
donald-pinckney: There are already details in core, but they were a bit wrong (like on changing args to activities). I've fixed those.

references/typescript/patterns.md
donald-pinckney: Added this context (and to Python too)

references/typescript/versioning.md
donald-pinckney: Ah, thats a great thing to call out. Fixed

references/typescript/patterns.md
donald-pinckney: yes, good call. improved this across typescript, core, and python.

references/typescript/gotchas.md
donald-pinckney: Agreed. improved this in core, python, and typescript

references/typescript/advanced-features.md
donald-pinckney: Got it, fixed.

donald-pinckney: Also a PR to fix the sample code: https://github.com/temporalio/samples-typescript/pull/451

references/typescript/testing.md
donald-pinckney: Sounds good, I'll leave it as-is for now.

PR #38, opened 2026-03-16
Add Go (supersedes other PR)

donald-pinckney:
## What was changed
Adds support for Go.

This PR will be merged instead of #6, which is the much older and way less complete version of Go. Once approved I will force-push these commits to the branch of #6, in case anyone was actually using #6 for testing.

### DIFF
--- a/references/core/determinism.md
+++ b/references/core/determinism.md
@@ -78,10 +78,11 @@ For a few simple cases, like timestamps, random values, UUIDs, etc. the Temporal
 ## SDK Protection Mechanisms
 Each Temporal SDK language provides a protection mechanism to make it easier to catch non-determinism errors earlier in development:
 
-- Python: The Python SDK runs workflows in a sandbox that intercepts and aborts non-deterministic calls at runtime.
+- Python: The Python SDK runs workflows in a sandbox that intercepts and aborts non-deterministic calls early at runtime. 
 - TypeScript: The TypeScript SDK runs workflows in an isolated V8 sandbox, intercepting many common sources of non-determinism and replacing them automatically with deterministic variants.
-- Go: The Go SDK has no runtime sandbox. Determinism is enforced by developer convention and the optional `workflowcheck` static analysis tool, which detects non-deterministic calls at compile time.
+- Go: The Go SDK has no runtime sandbox. Therefore, non-determinism bugs will never be immediately appararent, and are usually only observable during replay. The optional `workflowcheck` static analysis tool can be used to check for many sources of non-determinism at compile time.
 
+Regardless of which SDK you are using, it is your responsibility to ensure that workflow code does not contain sources of non-determinism. Use SDK-specific tools as well as replay tests for doing so.
 
 ## Detecting Non-Determinism
 


references/go/advanced-features.md
```diff
@@ -0,0 +1,180 @@
...
+    MaxConcurrentWorkflowTaskPollers: 4,
+
+    // Graceful shutdown timeout (default: 0)
+    WorkerStopTimeout: 30 * time.Second,
+})
+```
+
+Scale pollers based on task queue throughput. If you observe high schedule-to-start latency, increase the number of pollers or add more workers.
+
+## Sessions
```
mjameswh, line 127: I think it's pertinent to mention, but:
- We should call out limitations of that API (there are a few reasons we never ported that one to other SDKs; I'd suggest doing some research on this, there's been a lot said about that already).
- We should point out that it is _mostly_ equivalent to the worker-specific task queue pattern (again, I'll let you do some research).

references/go/data-handling.md
```diff
@@ -0,0 +1,183 @@
...
+```go
+converter.NewProtoPayloadConverter()
+```
+
+JSON protobuf:
+```go
+converter.NewProtoJSONPayloadConverter()
+```
+
+Both are included in the default data converter. SDK v1.26.0+ migrated from gogo/protobuf to google/protobuf. If you need backward compatibility with older payloads encoded with gogo, use the `LegacyTemporalProtoCompat` option.
```
mjameswh: I'd suggest adding the release date next to this version number. It seems likely that the very next agent's step after reading that sentence will be to search for that date to determine whether this backward compatibility concern is likely to apply to the user.

references/go/determinism.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Overview
+
+The Go SDK has NO runtime sandbox (unlike Python/TypeScript). Workflows must be deterministic for replay, and determinism is enforced entirely by developer convention and optional static analysis via the `workflowcheck` tool (see `references/go/determinism-protection.md`).
+
+## Why Determinism Matters: History Replay
+
+Temporal provides durable execution through **History Replay**. When a Worker restores workflow state, it re-executes workflow code from the beginning. This requires the code to be **deterministic**. See `references/core/determinism.md` for a deep explanation.
+
+## Forbidden Operations
```
mjameswh, line 11: There's a lot of duplication between this file and `determinism-protection.md`, and possibly with other files in the "shared" content at the root of the repo.

Not a huge concern, but that consumes tokens. You may want to revisit and optimize this at some point.

references/go/error-handling.md
```diff
@@ -0,0 +1,182 @@
...
+- **StartToCloseTimeout**: Max time for a single Activity Task Execution. Prefer this over ScheduleToCloseTimeout.
+- **ScheduleToCloseTimeout**: Total time including retries.
+- **ScheduleToStartTimeout**: Time an Activity Task can wait in the Task Queue before a Worker picks it up. Rarely needed.
+- **HeartbeatTimeout**: Max time between heartbeats. Required for long-running activities to detect failures.
+
+Either `StartToCloseTimeout` or `ScheduleToCloseTimeout` must be set.
+
+## Workflow Failure
+
+Returning any error from a workflow function fails the execution. Return `nil` for success.
```
mjameswh, line 149: This is a frequent gotcha for users. Go is the only SDK that fails execution by default on errors; other SDKs all retry by default on non-TemporalFailure errors. This might be useful to explicitly call out somewhere.

references/go/patterns.md
```diff
@@ -0,0 +1,527 @@
...
+    err := workflow.SetUpdateHandlerWithOptions(
+        ctx,
+        "add-item",
+        func(ctx workflow.Context, item string) (int, error) {
+            // Handler can mutate workflow state and return a value
+            items = append(items, item)
+            return len(items), nil
+        },
+        workflow.UpdateHandlerOptions{
+            Validator: func(ctx workflow.Context, item string) error {
```
mjameswh, line 104: Should point out that _validators_ can't mutate state or do anything blocking, similar to query handlers.

references/go/patterns.md
```diff
@@ -0,0 +1,527 @@
...
+```go
+func OrderWorkflow(ctx workflow.Context, order Order) (string, error) {
+    actCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+        StartToCloseTimeout: 5 * time.Minute,
+    })
+
+    var compensations []func(ctx workflow.Context) error
+
+    // Helper to run all compensations in reverse
+    runCompensations := func() {
```
mjameswh, line 366: In most cases, one would want to shield compensations with ctx.NewDisconnectedContext to ensure that compensations get run even if the workflow gets cancelled.

I'd suggest reordering this section after "Cancellation Handling" below, so that `NewDisconnectedContext` is introduced first.

references/go/patterns.md
```diff
@@ -0,0 +1,527 @@
...
+    // Cancel the timer when no longer needed
+    cancelTimer()
+
+    return "Timer fired", nil
+}
+```
+
+## Local Activities
+
+**Purpose**: Reduce latency for short, lightweight operations by skipping the task queue. ONLY use these when necessary for performance. Do NOT use these by default, as they are not durable and distributed.
```
mjameswh, line 520: We should also call out somewhere that calling multiple local activities in a row (i.e. in a same WFT, with nothing in-between to allow the WFT to complete), increases risks, because local activity completions only get persisted when the WFT completes.

donald-pinckney: Just added it to `core`.

references/go/patterns.md
donald-pinckney: I agree, and this is a particularly nasty gotcha that agents could easily fall into. For that reason, I've now mentioned it across all languages and in `core`.

references/go/determinism.md
donald-pinckney: Yup, this one is on my radar for being a remaining duplication offender.

references/core/determinism.md
```diff
@@ -80,6 +80,7 @@ Each Temporal SDK language provides a protection mechanism to make it easier to
 
 - Python: The Python SDK runs workflows in a sandbox that intercepts and aborts non-deterministic calls at runtime.
 - TypeScript: The TypeScript SDK runs workflows in an isolated V8 sandbox, intercepting many common sources of non-determinism and replacing them automatically with deterministic variants.
+- Go: The Go SDK has no runtime sandbox. Determinism is enforced by developer convention and the optional `workflowcheck` static analysis tool, which detects non-deterministic calls at compile time.
```
yuandrew: Not sure "developer convention" is the best term here, and also this makes it sound like `workflowcheck` is used to enforce determinism. I would make it more clear that `workflowcheck` is just a tool users can use to help avoid NDE.
```suggestion
- Go: The Go SDK has no runtime sandbox. Determinism is enforced by the SDK's replay-based execution model, which provides deterministic alternatives to standard library operations and detects non-deterministic
   divergence during replay. The optional `workflowcheck` static analysis tool can additionally help detect non-deterministic calls at compile time.
```

references/go/determinism.md
```diff
@@ -0,0 +1,51 @@
...
+- **Native goroutines** (`go func()`) -- use `workflow.Go()` instead
+- **Native channels** (`chan`, send, receive, `range` over channel) -- use `workflow.Channel` instead
+- **Native `select`** -- use `workflow.Selector` instead
+- **`time.Now()`** -- use `workflow.Now(ctx)` instead
+- **`time.Sleep()`** -- use `workflow.Sleep(ctx, duration)` instead
+- **`math/rand` global** (e.g., `rand.Intn()`) -- use `workflow.SideEffect` instead
+- **`crypto/rand.Reader`** -- use an activity instead
+- **`os.Stdin` / `os.Stdout` / `os.Stderr`** -- use `workflow.GetLogger(ctx)` for logging
+- **Map range iteration** (`for k, v := range myMap`) -- sort keys first, then iterate
+- **Mutating global variables** -- use local state or `workflow.SideEffect`
```
Quinn-With-Two-Ns, line 24: A unique Go specific issue is "Using an anonymous function as a local activity" since the name of the local activity will be non-deterministic.

Quinn-With-Two-Ns: The resolution is make all local activities proper functions.

references/go/observability.md
```diff
@@ -0,0 +1,153 @@
...
+
+logger := logur.LoggerToKV(logrusadapter.New(logrus.New()))
+c, err := client.Dial(client.Options{
+    Logger: logger,
+})
+```
+
+## Metrics
+
+Use the Tally library (`go.temporal.io/sdk/contrib/tally`) with Prometheus:
```
Quinn-With-Two-Ns, line 102: Why recommend tally here over otel?

donald-pinckney: otel was cut from scope of initial skill launch (https://github.com/temporalio/skill-temporal-developer/issues/23), since it was still in flux.

Tally & Prometheus is simple enough to include since its more narrowly scoped to metrics only. All other SDK languages in the skill have included metrics, but (for now) excluded tracing.

references/go/determinism.md
donald-pinckney: Great call. Added it.

references/go/data-handling.md
```diff
@@ -0,0 +1,183 @@
...
+2. `converter.NewByteSlicePayloadConverter()` -- `[]byte`
+3. `converter.NewProtoJSONPayloadConverter()` -- Protobuf messages as JSON
+4. `converter.NewProtoPayloadConverter()` -- Protobuf messages as binary
+5. `converter.NewJSONPayloadConverter()` -- anything JSON-serializable
+
+Structs must have exported fields to be serialized.
+
+## Custom Data Converter
+
+Implement the `converter.DataConverter` interface (`ToPayload`, `FromPayload`, `ToPayloads`, `FromPayloads`, `ToString`, `ToStrings`) or compose a new converter from existing ones.
```
yuandrew: More of a meta question to skills in general (as someone not super up-to-speed), this line reminds me that some of these docs are basically the same as the code itself, and now introduce a new place we have to keep in sync with the code itself, like if we happen to add a new method to this interface. I think someone could argue it's equally as fast to look at the code itself to see what methods the interface has.

Is this skill basically organizing the codebase in a AI digestible way so it doesn't need to dig through the code itself? Is adding information like this proven to be beneficial to the skill? Would the gain from this be comparable to if we ensured our docs were thorough and up to date?

donald-pinckney: Lots of questions here, I'll do what i can to break it down :).

**Q1: What value does a skill provide vs. docs vs. the SDK code itself?**

For reference, you can currently use these 2 MCP servers:
- Kapa: MCP server for docs access.
- context7: MCP server for SDK API search questions

So the question is, what value does a skill provide compared to those two MCP servers?

Primarily, its about providing a more tightly curated "garden path" to show to the agent, compared to the much broader in scope docs and SDK code. The aim is to make it both more digestable (= faster for the agent to index into the content that matters), and be able to *force* items into its context. 

As an example, consider the data converters. Just by loading the top level `SKILL.md`, the agent will (no choice) get links to `references/go/data-handling.md` in its context. If its doing some task where data serialization seems relevant, it would be highly likely to then load `references/go/data-handling.md`, since it knows it exists. If you are just using Kapa or context7, there is nothing in the agent's context that alerts the agent that Temporal has any concept of data handling / serialization / etc., so it might not decide to do an MCP call to lookup in the docs or SDK code about data conversion.

**Q2: How do we know this is beneficial to the agent?**

The only way we will know for sure is by doing extensive evaluations. I'm currently working on this, but if you want to help, the most helpful thing you can do is note down Temporal coding tasks that a coding agent performed poorly on. Given difficult tasks, we can then observe that the agent completes them successfully given the skill. Let me know if you know of any tricky tasks for agents!

**Q3: How do we keep skill in sync?**

This is discussed a bit in the PRD, but essentially I will track what SDK versions the skill was currently written for, and then in the future can diff changelogs, and identify and fix stale areas of the skill. I'd also like to investigate snipsync sorts of stuff for managing code snippets better.

donald-pinckney: That being said, I don't like this `Custom Data Converter` content thats here. I don't think its descriptive enough to really be helpful to the agent. Will fix.

donald-pinckney: Has been changed to a code sample.

yuandrew: Thanks for the answers! I'll take a closer look at the PRD as well

> if you want to help, the most helpful thing you can do is note down Temporal coding tasks that a coding agent performed poorly on.

sounds good, will keep an eye out and remember to give you feedback when I come across this 😃

references/core/determinism.md
donald-pinckney: I didn't take your edit, since i think its quite confusing to the agent to say that determinism is enforced by the SDK - that makes it sound like the agent isn't responsible for making sure workflow code is deterministic. But I agree, I also didn't like "developer convention" - it sounds too passive and vague. I just pushed an edit that should clarify things.

PR #39, opened 2026-03-16
Add .NET SDK support to temporal-developer skill

donald-pinckney:
## Summary
- Added 11 .NET reference files covering workflows, activities, patterns, determinism, error handling, testing, versioning, observability, data handling, gotchas, and advanced features (including DI)
- Updated SKILL.md with .NET trigger phrases and getting started entry
- Updated core/determinism.md SDK Protection Mechanisms with .NET entry
- .NET-specific content: Task determinism (no sandbox), CancellationToken patterns, dependency injection, safe builtin alternatives table

### DIFF
--- a/references/dotnet/advanced-features.md
+++ b/references/dotnet/advanced-features.md
@@ -33,8 +33,10 @@ await handle.DeleteAsync();
 ## Async Activity Completion
 
 For activities that complete asynchronously (e.g., human tasks, external callbacks).
+If you configure a `HeartbeatTimeout` on this activity, the external completer is responsible for sending heartbeats via the async handle.
+If you do NOT set a `HeartbeatTimeout`, no heartbeats are required.
 
-**Note:** If the external system can reliably Signal back with the result, consider using **signals** instead.
+**Note:** If the external system that completes the asynchronous action can reliably be trusted to do the task and Signal back with the result, and it doesn't need to Heartbeat or receive Cancellation, then consider using **signals** instead.
 
 ```csharp
 using Temporalio.Activities;
@@ -60,9 +62,13 @@ public async Task CompleteApprovalAsync(string requestId, bool approved)
 
     var handle = client.GetAsyncActivityHandle(taskToken);
 
+    // Optional: if a HeartbeatTimeout was set, you can periodically:
+    // await handle.HeartbeatAsync(progressDetails);
+
     if (approved)
         await handle.CompleteAsync("approved");
     else
+        // You can also fail or report cancellation via the handle
         await handle.FailAsync(new ApplicationFailureException("Rejected"));
 }
 ```
@@ -86,6 +92,36 @@ var worker = new TemporalWorker(
     .AddWorkflow<MyWorkflow>()
     .AddAllActivities(new MyActivities()));
 ```
+
+## Workflow Init Attribute
+
+Use `[WorkflowInit]` on a constructor to run initialization code when a workflow is first created.
+
+**Purpose:** Execute some setup code before signal/update happens or run is invoked.
+
+```csharp
+[Workflow]
+public class MyWorkflow
+{
+    private readonly string _initialValue;
+    private readonly List<string> _items = new();
+
+    [WorkflowInit]
+    public MyWorkflow(string initialValue)
+    {
+        _initialValue = initialValue;
+    }
+
+    [WorkflowRun]
+    public async Task<string> RunAsync(string initialValue)
+    {
+        // _initialValue and _items are already initialized
+        return _initialValue;
+    }
+}
+```
+
+Constructor and `[WorkflowRun]` method must have the same parameters with the same types. You cannot make blocking calls (activities, sleeps, etc.) from the constructor.
 
 ## Workflow Failure Exception Types
 
@@ -118,18 +154,24 @@ The .NET SDK supports dependency injection via the `Temporalio.Extensions.Hostin
 ```csharp
 using Temporalio.Extensions.Hosting;
 
-var host = Host.CreateDefaultBuilder(args)
-    .ConfigureServices(ctx =>
-        ctx.
-            AddScoped<IOrderRepository, OrderRepository>().
-            AddHostedTemporalWorker(
-                clientTargetHost: "localhost:7233",
-                clientNamespace: "default",
-                taskQueue: "my-task-queue").
-            AddScopedActivities<MyActivities>().
-            AddWorkflow<MyWorkflow>())
-    .Build();
-await host.RunAsync();
+public class Program
+{
+    public static async Task Main(string[] args)
+    {
+        var host = Host.CreateDefaultBuilder(args)
+            .ConfigureServices(ctx =>
+                ctx.
+                    AddScoped<IOrderRepository, OrderRepository>().
+                    AddHostedTemporalWorker(
+                        clientTargetHost: "localhost:7233",
+                        clientNamespace: "default",
+                        taskQueue: "my-task-queue").
+                    AddScopedActivities<MyActivities>().
+                    AddWorkflow<MyWorkflow>())
+            .Build();
+        await host.RunAsync();
+    }
+}
 ```
 
 ### Activity Dependency Injection
--- a/references/dotnet/advanced-features.md
+++ b/references/dotnet/advanced-features.md
@@ -1,0 +1,160 @@
+# .NET SDK Advanced Features
+
+## Schedules
+
+Create recurring workflow executions.
+
+```csharp
+using Temporalio.Client.Schedules;
+
+var scheduleId = "daily-report";
+await client.CreateScheduleAsync(
+    scheduleId,
+    new Schedule(
+        action: ScheduleActionStartWorkflow.Create(
+            (DailyReportWorkflow wf) => wf.RunAsync(),
+            new(id: "daily-report", taskQueue: "reports")),
+        spec: new ScheduleSpec
+        {
+            Intervals = new List<ScheduleIntervalSpec>
+            {
+                new(Every: TimeSpan.FromDays(1)),
+            },
+        }));
+
+// Manage schedules
+var handle = client.GetScheduleHandle(scheduleId);
+await handle.PauseAsync("Maintenance window");
+await handle.UnpauseAsync();
+await handle.TriggerAsync(); // Run immediately
+await handle.DeleteAsync();
+```
+
+## Async Activity Completion
+
+For activities that complete asynchronously (e.g., human tasks, external callbacks).
+
+**Note:** If the external system can reliably Signal back with the result, consider using **signals** instead.
+
+```csharp
+using Temporalio.Activities;
+using Temporalio.Client;
+
+[Activity]
+public async Task RequestApprovalAsync(string requestId)
+{
+    var taskToken = ActivityExecutionContext.Current.Info.TaskToken;
+
+    // Store task token for later completion (e.g., in database)
+    await StoreTaskTokenAsync(requestId, taskToken);
+
+    // Mark this activity as waiting for external completion
+    throw new CompleteAsyncException();
+}
+
+// Later, complete the activity from another process
+public async Task CompleteApprovalAsync(string requestId, bool approved)
+{
+    var client = await TemporalClient.ConnectAsync(new("localhost:7233"));
+    var taskToken = await GetTaskTokenAsync(requestId);
+
+    var handle = client.GetAsyncActivityHandle(taskToken);
+
+    if (approved)
+        await handle.CompleteAsync("approved");
+    else
+        await handle.FailAsync(new ApplicationFailureException("Rejected"));
+}
+```
+
+## Worker Tuning
+
+Configure worker performance settings.
+
+```csharp
+var worker = new TemporalWorker(
+    client,
+    new TemporalWorkerOptions("my-task-queue")
+    {
+        // Workflow task concurrency
+        MaxConcurrentWorkflowTasks = 100,
+        // Activity task concurrency
+        MaxConcurrentActivities = 100,
+        // Graceful shutdown timeout
+        GracefulShutdownTimeout = TimeSpan.FromSeconds(30),
+    }
+    .AddWorkflow<MyWorkflow>()
+    .AddAllActivities(new MyActivities()));
+```
+
+## Workflow Failure Exception Types
+
+Control which exceptions cause workflow failures vs workflow task retries.
+
+**Default behavior:** Only `ApplicationFailureException` fails a workflow. All other exceptions retry the workflow task forever (treated as bugs to fix with a code deployment).
+
+**Tip for testing:** Set `WorkflowFailureExceptionTypes` to include `Exception` so any unhandled exception fails the workflow immediately rather than retrying the workflow task forever. This surfaces bugs faster.
+
+```csharp
+var worker = new TemporalWorker(
+    client,
+    new TemporalWorkerOptions("my-task-queue")
+    {
+        // These exception types will fail the workflow execution (not just the task)
+        WorkflowFailureExceptionTypes = new[] { typeof(ArgumentException), typeof(InvalidOperationException) },
+    }
+    .AddWorkflow<MyWorkflow>()
+    .AddAllActivities(new MyActivities()));
+```
+
+## Dependency Injection
+
+The .NET SDK supports dependency injection via the `Temporalio.Extensions.Hosting` package, which integrates with .NET's generic host.
+
+### Worker as Generic Host
+
+```csharp
+using Temporalio.Extensions.Hosting;
+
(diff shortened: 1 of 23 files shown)


references/dotnet/observability.md
```diff
@@ -0,0 +1,96 @@
+# .NET SDK Observability
```
cecilphillip, line 1: Should reference [Temporalio.Extensions.OpenTelemetry](https://www.nuget.org/packages/Temporalio.Extensions.OpenTelemetry) and [Temporalio.Extensions.DiagnosticSource](https://www.nuget.org/packages/Temporalio.Extensions.DiagnosticSource) here

references/dotnet/observability.md
```diff
@@ -0,0 +1,96 @@
...
+### Enabling SDK Metrics
+
+```csharp
+using Temporalio.Extensions.OpenTelemetry;
+using OpenTelemetry;
+using OpenTelemetry.Metrics;
+
+// Configure OpenTelemetry metrics
+var meterProvider = Sdk.CreateMeterProviderBuilder()
+    .AddTemporalClientInstrumentation()
```
cecilphillip: Where does this come from? I don't see it in the SDK

references/dotnet/testing.md
```diff
@@ -0,0 +1,176 @@
...
+    {
+        var result = await env.Client.ExecuteWorkflowAsync(
+            (MyWorkflow wf) => wf.RunAsync("test"),
+            new(id: $"wf-{Guid.NewGuid()}", taskQueue: worker.Options.TaskQueue!));
+        Assert.Equal("mocked: test", result);
+    });
+}
+```
+
+**Note:** If the original activity method name ends with `Async`, the default activity name has `Async` trimmed off. For example, `MyActivityAsync` has default name `MyActivity`.
```
donald-pinckney: Confirm: is this true?

references/dotnet/observability.md
donald-pinckney: For the moment, language skills only support metrics via prometheus, which IIUC, these are not needed for. Will do these later as follow ups for issues #62 and #63, to get proper support for these concepts across all languages.

references/dotnet/observability.md
donald-pinckney: fixed, thanks!

references/dotnet/testing.md
cecilphillip: Confirmed
https://github.com/temporalio/sdk-dotnet/blob/b39d96d7032d4e2fbb52dec0a7e2556cc16cc5f4/src/Temporalio/Activities/ActivityDefinition.cs#L286-L334

donald-pinckney: Yes, mostly true. Made it a bit more precise.

partha-dutta-maersk (approved):

references/dotnet/advanced-features.md
```diff
@@ -0,0 +1,202 @@
...
+
+    // Mark this activity as waiting for external completion
+    throw new CompleteAsyncException();
+}
+
+// Later, complete the activity from another process
+public async Task CompleteApprovalAsync(string requestId, bool approved)
+{
+    var client = await TemporalClient.ConnectAsync(new("localhost:7233"));
+    var taskToken = await GetTaskTokenAsync(requestId);
```
jmaeagle99, line 62: nit: Make a symmetric comment that this is getting the task token from external storage or service, to help indicate this is customer defined.

references/dotnet/advanced-features.md
```diff
@@ -0,0 +1,202 @@
...
+                    AddWorkflow<MyWorkflow>())
+            .Build();
+        await host.RunAsync();
+    }
+}
+```
+
+### Activity Dependency Injection
+
+Activities registered with `AddScopedActivities<T>()` or `AddSingletonActivities<T>()` are created via DI, allowing constructor injection:
```
jmaeagle99: `AddTransientActivities` also exists.

references/dotnet/data-handling.md
```diff
@@ -0,0 +1,216 @@
+# .NET SDK Data Handling
+
+## Overview
+
+The .NET SDK uses data converters to serialize/deserialize workflow inputs, outputs, and activity parameters.
```
jmaeagle99, line 5: The behavior described here only applies to `PayloadConverter`. `DataConverter` is a higher level concept that combines payload converters, payload codecs, and failure converters to allow an overall conversion of raw values to payloads, transforming those payloads, and serializing errors. See https://github.com/temporalio/sdk-dotnet?tab=readme-ov-file#data-conversion

This is conceptually the same across all core based SDKs (Python, TypeScript, .NET, Ruby, Rust) but is different from what Go and Java consider a `DataConverter`. It's nuanced, but it needs to be documented correctly because developers and skills will try to incorrectly assume that all of "data conversion" works the same and is declared the same across all of the SDKs.

references/dotnet/data-handling.md
```diff
@@ -0,0 +1,216 @@
...
+var client = await TemporalClient.ConnectAsync(new("localhost:7233")
+{
+    DataConverter = DataConverter.Default with
+    {
+        PayloadCodec = new EncryptionCodec(),
+    },
+});
+```
+
+## Search Attributes
```
jmaeagle99, line 106: Not convinced that "search attributes" should be in a "data handling" document. Search attributes are not involved in handling user data in the commonly understood sense and they don't participate in codecs / encryption. Maybe this should be in a visibility document, especially if there are other visibility concepts.

references/dotnet/data-handling.md
```diff
@@ -0,0 +1,216 @@
...
+
+        return uniqueId.ToString();
+    }
+}
+```
+
+## Best Practices
+
+1. Use records or classes with `System.Text.Json` support for input/output
+2. Keep payloads small — see `references/core/gotchas.md` for limits
```
jmaeagle99, line 213: Make this a link instead of an inline code snippet.

references/dotnet/determinism-protection.md
```diff
@@ -0,0 +1,49 @@
+# .NET Determinism Protection
+
+## Overview
+
+The .NET SDK has no runtime sandbox. Determinism is enforced by **developer convention** and **runtime task detection**. Unlike the Python and TypeScript SDKs, the .NET SDK will not intercept or replace non-deterministic calls at compile time or import time. The SDK does provide a runtime `EventListener` that detects some invalid task scheduling, but catching all non-deterministic code requires following the rules below and testing, in particular replay tests (see `references/dotnet/testing.md`).
```
jmaeagle99, line 5: Same here about using links instead of code snippets.

references/dotnet/determinism-protection.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Runtime Task Detection
+
+By default, the .NET SDK enables an `EventListener` that monitors task events. When workflow code accidentally starts a task on the wrong scheduler (e.g., via `Task.Run`), an `InvalidWorkflowOperationException` is thrown. This "pauses" the workflow by failing the workflow task, which continually retries until the code is fixed.
```
jmaeagle99: I wouldn't say that "This \"pauses\" the workflow..." here since workflow pause is an actual concept, or is at least in the API if not exposed yet. Maybe something like "This causes the workflow task to fail, which will continuously retry..."

references/dotnet/determinism-protection.md
```diff
@@ -0,0 +1,49 @@
...
+- `System.Threading.Semaphore` / `SemaphoreSlim` / `Mutex` — use `Temporalio.Workflows.Semaphore` / `Mutex`.
+
+**Be wary of:**
+- Third-party libraries that implicitly use `TaskScheduler.Default`
+- `Dataflow` blocks and similar concurrency libraries with hidden default scheduler usage
+
+## Best Practices
+
+1. **Always use `Workflow.*` alternatives** for Task operations in workflows
+2. **Enable the `EventListener`** (default) — it catches mistakes at runtime
```
jmaeagle99: Does this need to be called out as a best practice action if its already the default? There is no action to take here. If it needs to be called out, maybe "don't disable the event listener".

references/dotnet/determinism.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Overview
+
+The .NET SDK has NO runtime sandbox (unlike Python/TypeScript). Workflows must be deterministic for replay, and determinism is enforced by developer convention and runtime task detection via an `EventListener` (see `references/dotnet/determinism-protection.md`).
+
+## Why Determinism Matters: History Replay
+
+Temporal provides durable execution through **History Replay**. When a Worker restores workflow state, it re-executes workflow code from the beginning. This requires the code to be **deterministic**. See `references/core/determinism.md` for a deep explanation.
+
+## Forbidden Operations
```
jmaeagle99: I think it should be clearly called out these are "forbidden" for workflows, but are fine to use in activities. Maybe the line at the end of the section is enough, but I sat here reading through the list to eventually see "oh, so I can use these, just should only do so in activities".

references/dotnet/dotnet.md
```diff
@@ -0,0 +1,193 @@
+# Temporal .NET SDK Reference
+
+## Overview
+
+The Temporal .NET SDK provides a high-performance, type-safe approach to building durable workflows using C# and .NET. Workflows use attributes (`[Workflow]`, `[WorkflowRun]`) and lambda expressions for type-safe invocations. .NET 6.0+ required.
```
jmaeagle99: Currently, the SDK supports .NET Framework 4.6.2+ and .NET Core 3.1+ (which includes .NET 5+). The .NET Core version min might change soon. Officially, Microsoft only supports .NET 8+, and even .NET 8 will be EOL at the end of this year.

references/dotnet/error-handling.md
```diff
@@ -0,0 +1,157 @@
...
+    [WorkflowRun]
+    public async Task<string> RunAsync()
+    {
+        try
+        {
+            return await Workflow.ExecuteActivityAsync(
+                (MyActivities a) => a.RiskyActivityAsync(),
+                new() { StartToCloseTimeout = TimeSpan.FromMinutes(5) });
+        }
+        catch (ActivityFailureException ex)
```
jmaeagle99:
```suggestion
        catch (ActivityFailureException ex) when (!TemporalException.IsCanceledException(ex))
```
Workflows should honor workflow cancellation and not treat it as a workflow failure.

references/dotnet/error-handling.md
```diff
@@ -0,0 +1,157 @@
...
+                ScheduleToCloseTimeout = TimeSpan.FromMinutes(30),  // Including retries
+                HeartbeatTimeout = TimeSpan.FromMinutes(2),         // Between heartbeats
+            });
+    }
+}
+```
+
+## Workflow Failure
+
+**Critical .NET behavior:** Only `ApplicationFailureException` will fail a workflow. All other exceptions (including standard .NET exceptions like `NullReferenceException`, `KeyNotFoundException`, etc.) will **retry the workflow task** indefinitely. This is by design — those are treated as bugs to be fixed with a code deployment, not reasons for the workflow to fail.
```
jmaeagle99: Same point as https://github.com/temporalio/skill-temporal-developer/pull/39/changes#r3082703929

references/dotnet/advanced-features.md
```diff
@@ -0,0 +1,202 @@
...
+}
+```
+
+Constructor and `[WorkflowRun]` method must have the same parameters with the same types. You cannot make blocking calls (activities, sleeps, etc.) from the constructor.
+
+## Workflow Failure Exception Types
+
+Control which exceptions cause workflow failures vs workflow task retries.
+
+**Default behavior:** Only `ApplicationFailureException` fails a workflow. All other exceptions retry the workflow task forever (treated as bugs to fix with a code deployment).
```
jmaeagle99, line 131: > Only `ApplicationFailureException` fails a workflow

Should this be called out as being SDK specific? I don't think any other SDK has this behavior except if non-retryable is set to true.

jmaeagle99 (commented):
Still reviewing but want to submit some feedback.

references/dotnet/patterns.md
```diff
@@ -0,0 +1,484 @@
...
+        {
+            Workflow.Logger.LogError(ex, "Order failed, running compensations");
+            compensations.Reverse();
+            foreach (var compensate in compensations)
+            {
+                try { await compensate(); }
+                catch (Exception compErr)
+                {
+                    Workflow.Logger.LogError(compErr, "Compensation failed");
+                }
```
jmaeagle99, line 324: nit: Should this catch reraise so that latter compensations don't run if the current one failed?

references/dotnet/patterns.md
```diff
@@ -0,0 +1,484 @@
...
+            Workflow.Logger.LogError(e, "Cancellation occurred, performing cleanup");
+            // Use a detached cancellation token for cleanup since Workflow.CancellationToken
+            // is now cancelled
+            using var detachedCancelSource = new CancellationTokenSource();
+            await Workflow.ExecuteActivityAsync(
+                (MyActivities a) => a.CleanupAsync(),
+                new()
+                {
+                    StartToCloseTimeout = TimeSpan.FromMinutes(5),
+                    CancellationToken = detachedCancelSource.Token,
```
jmaeagle99:
```suggestion
                    CancellationToken = CancellationToken.None,
```
instead of creating a `CancellationTokenSource` that doesn't do anything. If the cleanup does need to be cancelled somehow (which isn't demonstrated in this example), then using `CancellationTokenSource` would be more appropriate.

references/dotnet/patterns.md
```diff
@@ -0,0 +1,484 @@
...
+- **Checkpointing** — Save progress periodically
+
+```csharp
+[Activity]
+public async Task<string> ProcessLargeFileAsync(string filePath)
+{
+    var info = ActivityExecutionContext.Current.Info;
+    // Get heartbeat details from previous attempt (if any)
+    var startLine = info.HeartbeatDetails.Count > 0
+        ? await info.HeartbeatDetails.ElementAtAsync<int>(0)
```
jmaeagle99:
```suggestion
        ? await info.HeartbeatDetailAtAsync<int>(0)
```

references/dotnet/data-handling.md
donald-pinckney: That's something I need to change in a ton of places skill-wide, will do separately.

references/dotnet/data-handling.md
donald-pinckney: I agree. There is currently a Search Attributes section of observability.md that just links to the data handling file. I'm thinking of moving search attributes there completely, but thats a cross-language change that I'll do in a later PR.

references/dotnet/determinism-protection.md
donald-pinckney: Agreed, that "pause" is too imprecise. Good catch.

references/dotnet/determinism.md
donald-pinckney: Thats a very good call. Important to have that distinction, and an easy change to make, so I've made it language-wide right now.

references/dotnet/patterns.md
donald-pinckney: I think these docs are also out of date then:

https://docs.temporal.io/develop/dotnet/workflows/cancellation#cancellation

donald-pinckney: Docs PR: https://github.com/temporalio/documentation/pull/4446

references/dotnet/patterns.md
donald-pinckney: Maybe, but IMO let's not go too deep on exactly the right compensation semantics. I'd argue it may be customer / situation dependent, so let's leave it on the simpler side. (IMO we would be better off providing a small API that implements all the edge-cases of saga).

references/dotnet/advanced-features.md
donald-pinckney: I think its fine to not specifically call out it being SDK specific, the agent is reading this file because its working within .NET code. Honestly, all the differences of error handling between all the SDKs is fairly confusing even to me, so I think a future pass over all of them checking for and clarifying semantics would be good, but thats out of scope here.

references/dotnet/advanced-features.md
donald-pinckney: Yup, good idea. Added it to the other languages too.

references/dotnet/error-handling.md
donald-pinckney: Fixed, and also in Java and Python.

references/dotnet/data-handling.md
donald-pinckney: Yes, I definitely agree. This is already an issue in Python and TypeScript as well. Since this is a bit of a bigger rewrite, I'm going to defer this to a follow-up PR that fixes it across all 3 languages. Issue for it has been created here: https://github.com/temporalio/skill-temporal-developer/issues/86

PR #41, opened 2026-03-16
Add Ruby SDK support

donald-pinckney:
## Summary
- Add 11 Ruby reference files (~2100 lines) covering all major Temporal SDK features: workflows, activities, signals/queries/updates, child workflows, versioning, testing, error handling, data conversion, observability, determinism protection, and advanced features
- Update SKILL.md to include Ruby in description, overview, and getting started guide
- Add Ruby entry to core/determinism.md SDK Protection Mechanisms section

## Details
The Ruby SDK (`temporalio` gem, GA) uses a fundamentally different determinism approach from Python (sandbox) and TypeScript (V8 isolate):
- **Illegal Call Tracing** via Ruby's `TracePoint` to detect forbidden method calls at runtime
- **Durable Fiber Scheduler** to make fiber operations deterministic

Ruby-specific sections added that don't exist in Python/TypeScript:
- Cancellation (Token-based) — uses `Temporalio::Cancellation` objects
- ActiveModel Integration — Ruby equivalent of Pydantic
- Converter Hints — `workflow_arg_hint`/`workflow_result_hint` for type guidance
- Illegal Call Tracing Gotchas — third-party gems triggering NondeterminismError
- Activity Concurrency and Executors — ThreadPool configuration
- Rails Integration — ActiveRecord, Zeitwerk, forking considerations

## Test plan
- [ ] Verify Ruby reference files are loaded correctly by the skill
- [ ] Test with a sample Ruby Temporal project to validate code examples
- [ ] Review alignment with Python/TypeScript content structure

### DIFF
--- a/references/ruby/advanced-features.md
+++ b/references/ruby/advanced-features.md
@@ -1,0 +1,191 @@
+# Ruby SDK Advanced Features
+
+## Schedules
+
+Create recurring workflow executions with `Temporalio::Client::Schedule`.
+
+```ruby
+require 'temporalio/client'
+
+# Create a schedule
+schedule_id = 'daily-report'
+client.create_schedule(
+  schedule_id,
+  Temporalio::Client::Schedule.new(
+    action: Temporalio::Client::Schedule::Action::StartWorkflow.new(
+      DailyReportWorkflow,
+      id: 'daily-report',
+      task_queue: 'reports'
+    ),
+    spec: Temporalio::Client::Schedule::Spec.new(
+      intervals: [
+        Temporalio::Client::Schedule::Spec::Interval.new(every: 86_400) # 1 day in seconds
+      ]
+    )
+  )
+)
+
+# Manage schedules
+handle = client.schedule_handle(schedule_id)
+handle.pause(note: 'Maintenance window')
+handle.unpause
+handle.trigger
+handle.delete
+```
+
+## Async Activity Completion
+
+For activities that complete asynchronously (e.g., human tasks, external callbacks).
+
+```ruby
+class RequestApproval < Temporalio::Activity::Definition
+  def execute(request_id)
+    # Get task token for async completion
+    task_token = Temporalio::Activity::Context.current.info.task_token
+
+    # Store task token for later completion (e.g., in database)
+    store_task_token(request_id, task_token)
+
+    # Signal that this activity completes asynchronously
+    Temporalio::Activity::Context.current.raise_complete_async
+  end
+end
+
+# Later, complete the activity from another process
+client = Temporalio::Client.connect('localhost:7233')
+task_token = get_task_token(request_id)
+handle = client.async_activity_handle(task_token: task_token)
+
+if approved
+  handle.complete('approved')
+else
+  handle.fail(Temporalio::Error::ApplicationError.new('Rejected'))
+end
+```
+
+If you configure a `heartbeat_timeout:` on the activity, the external completer is responsible for sending heartbeats via the async handle. If you do NOT set a `heartbeat_timeout`, no heartbeats are required.
+
+## Worker Tuning
+
+Configure worker performance settings.
+
+```ruby
+worker = Temporalio::Worker.new(
+  client: client,
+  task_queue: 'my-queue',
+  workflows: [MyWorkflow],
+  activities: [MyActivity],
+  max_concurrent_workflow_tasks: 100,
+  max_concurrent_activities: 100
+)
+worker.run
+```
+
+## Workflow Failure Exception Types
+
+Control which exceptions cause workflow failure vs workflow task failure (which Temporal retries automatically).
+
+### Per-Workflow Configuration
+
+```ruby
+class MyWorkflow < Temporalio::Workflow::Definition
+  # Class method approach
+  def self.workflow_failure_exception_type
+    MyCustomError
+  end
+
+  def execute
+    raise MyCustomError, 'This fails the workflow, not just the task'
+  end
+end
+```
+
+### Worker-Level Configuration
+
+```ruby
+Temporalio::Worker.new(
+  client: client,
+  task_queue: 'my-queue',
+  workflows: [MyWorkflow],
+  workflow_failure_exception_types: [MyCustomError]
+)
+```
+
+**Tips:**
+- Set to `[Exception]` in tests so any unhandled exception fails the workflow immediately rather than retrying the workflow task forever. Surfaces bugs faster.
+- Include `Temporalio::Workflow::NondeterminismError` to fail the workflow instead of leaving it in a retrying state on non-determinism errors.
+
+## Activity Concurrency and Executors
+
+Ruby uses `Temporalio::Worker::ActivityExecutor::ThreadPool` by default. Activities run in a thread pool.
+
+```ruby
+# Default: activities run in thread pool
+worker = Temporalio::Worker.new(
+  client: client,
+  task_queue: 'my-queue',
+  workflows: [MyWorkflow],
+  activities: [MyActivity],
+  activity_executors: {
+    default: Temporalio::Worker::ActivityExecutor::ThreadPool.new(max_threads: 20)
+  }
+)
+```
+
+Fiber-based execution is also possible for IO-bound activities using Ruby's fiber scheduler.
+
+## Rails Integration
+
+### ActiveRecord Considerations
+
+Never pass ActiveRecord models directly to Temporal workflows or activities. Serialize to plain data structures.
+
+```ruby
+# BAD - Passing AR model
+client.execute_workflow(
+  ProcessOrderWorkflow,
+  Order.find(42),  # Don't pass AR objects!
+  id: 'order-42',
+  task_queue: 'orders'
+)
+
+# GOOD - Pass serializable data
+client.execute_workflow(
+  ProcessOrderWorkflow,
+  { id: 42, total: order.total, status: order.status },
+  id: 'order-42',
+  task_queue: 'orders'
+)
+```
+
+### Zeitwerk and Autoloading
+
+Rails uses Zeitwerk for autoloading. Workflow and activity classes must be loadable by Zeitwerk or explicitly required.
+
+```ruby
+# In config/initializers/temporal.rb or similar
+# Eager load Temporal classes so they're available to the worker
+Rails.application.config.after_initialize do
+  Dir[Rails.root.join('app/workflows/**/*.rb')].each { |f| require f }
+  Dir[Rails.root.join('app/activities/**/*.rb')].each { |f| require f }
+end
+```
+
+### Forking Considerations
+
+If using a forking server (Puma, Unicorn), workers must be created **after** the fork. Connections established before fork are not safe to share across processes.
+
+```ruby
+# In Puma config (puma.rb)
+on_worker_boot do
+  # Create Temporal client and worker AFTER fork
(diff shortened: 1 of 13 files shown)


jb08:
Thank you for doing this! As a ruby sdk user, very interested in this MR. Any update on ETA?

hughevans:
Looks good!

references/ruby/advanced-features.md
```diff
@@ -0,0 +1,191 @@
...
+end
+```
+
+### Forking Considerations
+
+If using a forking server (Puma, Unicorn), workers must be created **after** the fork. Connections established before fork are not safe to share across processes.
+
+```ruby
+# In Puma config (puma.rb)
+on_worker_boot do
```
bdewater-thatch: This has been [renamed in Puma 7](https://github.com/puma/puma/blob/4196803/docs/7.0-Upgrade.md?plain=1#L40).
```suggestion
before_worker_boot do
```

references/ruby/advanced-features.md
```diff
@@ -0,0 +1,191 @@
...
+  id: 'order-42',
+  task_queue: 'orders'
+)
+```
+
+### Zeitwerk and Autoloading
+
+Rails uses Zeitwerk for autoloading. Workflow and activity classes must be loadable by Zeitwerk or explicitly required.
+
+```ruby
```
bdewater-thatch, line 212: Between the line above and this example it's not clear if the example code is required or not? "Loadable by Zeitwerk" seems to imply it should be autoloadable/reloadable in development without additional work, assuming the conventions are followed. Is that right?

In my experience both humans and and LLMs will default to using the example if it's not clear it is not required.

references/ruby/gotchas.md
```diff
@@ -0,0 +1,249 @@
...
+# GOOD - Regular heartbeats with progress
+class ProcessLargeFile < Temporalio::Activity::Definition
+  def execute(path)
+    File.readlines(path).each_slice(1000).with_index do |chunk, i|
+      Temporalio::Activity::Context.current.heartbeat("Processing chunk #{i}")
+      process(chunk)
+    end
+  end
+end
+```
```
bdewater-thatch: This example seems a lot like the one above.

Personally I'd like to see an example that allows me to compare Temporal activities with solutions like `job-iteration` or [Active Job Continuation](https://api.rubyonrails.org/classes/ActiveJob/Continuation.html). Claude came up with the following:
```rb
class ProcessAllUsers < Temporalio::Activity::Definition
  def execute
    context = Temporalio::Activity::Context.current
    cursor = context.info.heartbeat_details.first # nil on first attempt

    # `start` argument below is inclusive
    User.find_each(start: cursor) do |user|
      context.cancellation.check!
      process(user)
      context.heartbeat(user.id.succ) # next id to start from
    end
  end
end
```

references/ruby/gotchas.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Heartbeating
+
+### Forgetting to Heartbeat Long Activities
+
+```ruby
+# BAD - No heartbeat, can't detect stuck activities
+class ProcessLargeFile < Temporalio::Activity::Definition
+  def execute(path)
+    File.readlines(path).each_slice(1000) do |chunk|
```
bdewater-thatch:
```suggestion
    File.foreach(path).each_slice(1000) do |chunk|
```
Same for the other example below. See https://felipeelias.github.io/ruby/2017/01/02/fast-file-processing-ruby.html

chadxz:
Yes please

references/ruby/ruby.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Run the workflow:** Run `ruby execute_workflow.rb`. Should output: `Result: Hello, my name!`.
+
+## Key Concepts
+
+### Workflow Definition
+- Subclass `Temporalio::Workflow::Definition`
+- Define `def execute(args)` as the entry point
+- Use `Temporalio::Workflow.execute_activity` to call activities
+- Use signals, queries, and updates via class-level DSL methods
```
chris-olszewski:
```suggestion
- Define signals, queries, and updates via class-level DSL methods
```

references/ruby/gotchas.md
```diff
@@ -0,0 +1,249 @@
...
+class GoodWorkflow < Temporalio::Workflow::Definition
+  def execute
+    Temporalio::Workflow.sleep(60)  # Deterministic, durable timer
+  end
+end
+```
+
+**Why this matters:** `Kernel.sleep` uses the system clock, which differs between original execution and replay. `Temporalio::Workflow.sleep` creates a durable timer in the event history, ensuring consistent behavior during replay.
+
+## Illegal Call Tracing Gotchas
```
chris-olszewski, line 210: Unsure if it deserves a special callout, but I have seen misuse of Rail's ActiveRecord thinking operations won't trigger I/O. Could probably directly copy the callout in our rails sample: https://github.com/temporalio/samples-ruby/tree/main/rails_app#database-models-vs-temporal-models

references/ruby/testing.md
```diff
@@ -0,0 +1,234 @@
...
+
+Use `WorkflowReplayer` to verify that workflow code changes remain compatible with existing histories:
+
+```ruby
+require 'temporalio/worker/workflow_replayer'
+require 'temporalio/workflow_history'
+
+class ReplayTest < Minitest::Test
+  def test_replay_from_json
+    json = File.read('test/fixtures/my_workflow_history.json')
```
chris-olszewski, line 189: Would showing how to construct this JSON file be helpful?

references/ruby/patterns.md
```diff
@@ -0,0 +1,396 @@
...
+      end
+      raise
+    end
+  end
+end
+```
+
+## Cancellation (Token-based)
+
+Ruby uses `Temporalio::Cancellation` tokens instead of asyncio cancellation.
```
chris-olszewski: `asyncio` seems to be in reference to Python
```suggestion
Ruby uses `Temporalio::Cancellation` tokens.
```

references/ruby/observability.md
```diff
@@ -0,0 +1,62 @@
+# Ruby SDK Observability
```
chris-olszewski, line 1: Might be worth documenting the OTEL interceptor we provide: https://github.com/temporalio/samples-ruby/blob/fc306b7c72157169ba22112758d58c6becd59eec/open_telemetry/worker.rb#L27

references/ruby/determinism-protection.md
```diff
@@ -0,0 +1,157 @@
...
+- `Timeout.timeout` -- use `Temporalio::Workflow.sleep` with cancellation
+
+Note: `Time.new('2000-12-31')` with arguments IS deterministic and allowed. Only `Time.now` (wall-clock) is forbidden.
+
+## Durable Fiber Scheduler
+
+The SDK provides a custom `Fiber::Scheduler` implementation that makes fiber primitives deterministic within the workflow.
+
+- `Kernel.sleep` and `Mutex` technically work under the durable scheduler but are **disabled by default** via illegal call tracing to prevent accidental misuse.
+- `Temporalio::Workflow.sleep` and `Temporalio::Workflow.wait_condition` are the correct alternatives.
```
chris-olszewski: We do offer [`Temporalio::Workflow::Mutex`](https://ruby.temporal.io/Temporalio/Workflow.html) which is perfectly safe to use if a user truly wants a mutex and not a `wait_condition`

references/ruby/determinism-protection.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+Hash format:
+
+- `{ 'ClassName' => :all }` -- block all methods on the class.
+- `{ 'ClassName' => { method_name: true } }` -- block specific methods.
+
+## Common Issues
+
+### Third-party gems triggering NondeterminismError
```
chris-olszewski, line 97: Should the first fix be to attempt to move the call into an activity? We should only be suggesting these escape hatches after there has been verification that the call will not introduce non-determinism.

chris-olszewski (approved):
Overall looks good. Just a few suggestions

bdewater-thatch (commented):
Saw this mentioned on the Slack. Long time Ruby/Rails user, still a bit new to Temporal, so reviewing through that lens :)

references/ruby/gotchas.md
donald-pinckney: Yes, they are similar, because they are good/bad paired examples to show the model how to use `heartbeat(...)` correctly.

job-iteration / Active Job Continuation are interesting, but probably for follow-up PR(s).

references/ruby/gotchas.md
```diff
@@ -0,0 +1,249 @@
...
+    end
+  end
+end
+```
+
+```ruby
+# GOOD - Regular heartbeats with progress
+class ProcessLargeFile < Temporalio::Activity::Definition
+  def execute(path)
+    File.readlines(path).each_slice(1000).with_index do |chunk, i|
```
donald-pinckney:
```suggestion
    File.foreach(path).each_slice(1000).with_index do |chunk, i|
```

references/ruby/advanced-features.md
```diff
@@ -0,0 +1,191 @@
...
+  ProcessOrderWorkflow,
+  { id: 42, total: order.total, status: order.status },
+  id: 'order-42',
+  task_queue: 'orders'
+)
+```
+
+### Zeitwerk and Autoloading
+
+Rails uses Zeitwerk for autoloading. Workflow and activity classes must be loadable by Zeitwerk or explicitly required.
```
chris-olszewski:
```suggestion
Rails autoloading can result in unexpected I/O during replay. `config.eager_load` must be enabled or Workflows must explicitly require code dependencies before they are executed.
```

references/ruby/advanced-features.md
```diff
@@ -0,0 +1,191 @@
...
+
+```ruby
+# In config/initializers/temporal.rb or similar
+# Eager load Temporal classes so they're available to the worker
+Rails.application.config.after_initialize do
+  Dir[Rails.root.join('app/workflows/**/*.rb')].each { |f| require f }
+  Dir[Rails.root.join('app/activities/**/*.rb')].each { |f| require f }
+end
```
chris-olszewski: It might be more beneficial to show this happening at the worker level like we do in the sample: https://github.com/temporalio/samples-ruby/blob/main/rails_app/script/temporal_worker.rb#L3

references/ruby/advanced-features.md
donald-pinckney: Thanks for calling this out, it definitely wasn't clear, and has been imporved

references/ruby/observability.md
donald-pinckney: yes, but in a follow up (since it has to get done for the other langs too!)

references/ruby/determinism-protection.md
donald-pinckney: I think this Durable Fiber Scheduler section is not really providing any new / useful info, so I've removed it, and mentioned mutex above.

references/ruby/testing.md
donald-pinckney: Yes, but frankly other languages lack this, so I'm going to pull it into a separate issue and do as a follow-up.

references/ruby/determinism-protection.md
donald-pinckney: Yes, re-worked this section

PR #42, opened 2026-03-16
Add Java SDK support

donald-pinckney:
## Summary
- Add 11 Java SDK reference files covering patterns, determinism, error handling, testing, versioning, data handling, observability, and advanced features
- Update SKILL.md to include Java trigger phrases and language references
- Update core/determinism.md SDK Protection Mechanisms with Java entry (no sandbox, convention-based enforcement)

## Details
Java SDK support follows the same code-first, minimal prose style as Python/TypeScript references. Key Java-specific aspects:
- **No sandbox**: Determinism enforced via developer conventions + runtime replay detection (`NonDeterministicException`)
- **Interface + Implementation pattern**: `@WorkflowInterface`/`@WorkflowMethod`, `@ActivityInterface`/`@ActivityMethod`
- **`Workflow.getVersion()`** for patching (not `patched()`)
- **Cooperative threading**: One workflow thread at a time under global lock
- **Jackson** as default JSON serializer, **Micrometer** for metrics, **SLF4J** for logging

### DIFF
--- a/references/java/determinism-protection.md
+++ b/references/java/determinism-protection.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The Java SDK has **no sandbox**. Unlike the Python SDK (which blocks non-deterministic operations at runtime) or the TypeScript SDK (which replaces non-deterministic functions with deterministic variants), Java relies on developer conventions and runtime replay detection to enforce determinism. A static analysis tool (`temporal-workflowcheck`) is available in beta.
+The Java SDK has **no sandbox** (only Python and TypeScript have sandboxing). Java relies on developer conventions and runtime replay detection to enforce determinism. A static analysis tool (`temporal-workflowcheck`) is available in beta.
 
 ## Forbidden Operations
 
--- a/references/java/determinism.md
+++ b/references/java/determinism.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The Java SDK has **no sandbox**. Unlike the Python SDK (which uses a sandbox to block non-deterministic calls) or the TypeScript SDK (which uses V8 isolation to replace functions), the Java SDK relies on developer conventions to enforce determinism. The SDK provides `Workflow.*` APIs as safe replacements for common non-deterministic operations. A static analysis tool (`temporal-workflowcheck`, beta) can catch violations at build time — see `references/java/determinism-protection.md`.
+The Java SDK has **no sandbox** (only Python and TypeScript have sandboxing). The Java SDK relies on developer conventions to enforce determinism. The SDK provides `Workflow.*` APIs as safe replacements for common non-deterministic operations. A static analysis tool (`temporal-workflowcheck`, beta) can catch violations at build time — see `references/java/determinism-protection.md`.
 
 ## Why Determinism Matters: History Replay
 
--- a/references/java/java.md
+++ b/references/java/java.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-The Temporal Java SDK (`io.temporal:temporal-sdk`) uses an interface + implementation pattern for both Workflows and Activities. Java 8+ required; Java 21+ strongly recommended for virtual thread support. Unlike the Python and TypeScript SDKs, the Java SDK does **not** have a sandbox for determinism protection -- developers must ensure workflow code is deterministic themselves.
+The Temporal Java SDK (`io.temporal:temporal-sdk`) uses an interface + implementation pattern for both Workflows and Activities. Java 8+ required; Java 21+ strongly recommended for virtual thread support.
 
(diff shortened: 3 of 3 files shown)


references/java/java.md
```diff
@@ -0,0 +1,248 @@
...
+        System.out.println("Result: " + result);
+    }
+}
+```
+
+**Run the workflow:** Run `Starter.main()`. Should output: `Result: Hello, my name!`.
+
+## Key Concepts
+
+### Workflow Definition
```
Quinn-With-Two-Ns, line 163: What about `@WorkflowInit`? I think it is worth documenting

donald-pinckney: Good call. It is in Python, it should be here as well.

brianstrauch: Added to the Python skill [here](https://github.com/temporalio/skill-temporal-developer/pull/57) and also for the Java skill in this PR

references/java/java.md
```diff
@@ -0,0 +1,249 @@
...
+- Annotate interface with `@WorkflowInterface`
+- Use `@WorkflowInit` on the constructor to receive workflow input and initialize state before `@WorkflowMethod` or any signal/update handler is invoked
```
brianstrauch:
```suggestion
- Put any state initialization logic in the workflow constructor to guarantee that it happens before signals/updates arrive. If your state initialization logic requires the workflow parameters, then add the `@WorkflowInit` decorator and parameters to your constructor.
```

references/java/java.md
```diff
@@ -0,0 +1,249 @@
+# Temporal Java SDK Reference
+
+## Overview
+
+The Temporal Java SDK (`io.temporal:temporal-sdk`) uses an interface + implementation pattern for both Workflows and Activities. Java 8+ required. Unlike the Python and TypeScript SDKs, the Java SDK does **not** have a sandbox for determinism protection -- developers must ensure workflow code is deterministic themselves.
```
mjameswh: > Java 8+ required

True, but we should mention that Java 21+ is recommended, because of virtual thread support. ThatWe don't wouldn't want agents to bootstrap new Temporal projects with Java 8.

references/java/determinism.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Why Determinism Matters: History Replay
+
+Temporal provides durable execution through **History Replay**. When a Worker needs to restore workflow state (after a crash, cache eviction, or to continue after a long timer), it re-executes the workflow code from the beginning, which requires the workflow code to be **deterministic**.
+
+## SDK Protection
+
+Java workflow code runs in a cooperative threading model where only one workflow thread executes at a time under a global lock. The SDK does not intercept or block non-deterministic calls at runtime. If you call a forbidden operation, it will silently succeed during the initial execution but cause a `NonDeterministicException` when the workflow is replayed.
+
+Use `temporal-workflowcheck` (static analysis, beta) to catch violations at build time, and `WorkflowReplayer` to verify replay compatibility in tests.
```
mjameswh: I'd tune this sentence down a little bit. workflowcheck and replayer may _help_ uncover _some_ determinism violations, but agents shouldn't rely only on that to determine that a workflow is safe.

references/java/determinism-protection.md
```diff
@@ -0,0 +1,83 @@
+# Java Determinism Protection
+
+## Overview
+
+The Java SDK has **no sandbox**. Unlike the Python SDK (which blocks non-deterministic operations at runtime) or the TypeScript SDK (which replaces non-deterministic functions with deterministic variants), Java relies on developer conventions and runtime replay detection to enforce determinism. A static analysis tool (`temporal-workflowcheck`) is available in beta.
```
mjameswh: > Unlike the Python SDK (which blocks non-deterministic operations at runtime) or the TypeScript SDK (which replaces non-deterministic functions with deterministic variants),

This sentence is repeated in many places. Though it may be useful to a certain extent to mention that Java is "unlike some other SDKs ...", I really don't think it's pertinent to repeate that information that often, and definitely not pertinent to mention the way that Python and TS sandboxes works here.

That would also apply to other SDKs.

In short, TS and Python are the only two SDKs that benefit from sandboxing.

references/java/java.md
donald-pinckney: nice, good catch.

references/java/determinism.md
donald-pinckney: agreed!

references/java/determinism-protection.md
donald-pinckney: Yeah I agree, too repetitive. I've fixed the Java ones here, and the Go ones here: https://github.com/temporalio/skill-temporal-developer/pull/67

Will do the same in .NET.

PR #53, opened 2026-03-23
Add Nexus to Python Skill

Quinn-With-Two-Ns:
## What was changed

Add Nexus to Python Skill

## Why?

So Python developers using this skill get a easier experience using Nexus.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/references/python/nexus.md
+++ b/references/python/nexus.md
@@ -196,7 +196,7 @@ raise nexusrpc.OperationError(
 raise nexusrpc.HandlerError(
     "Service temporarily unavailable",
     type=nexusrpc.HandlerErrorType.INTERNAL,
-    retryable=True,
+    retryable_override=True,
 )
 ```
 


references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+A reverse proxy that routes requests from a caller Namespace to a handler Namespace and Task Queue. Endpoints decouple callers from handlers — the caller only knows the endpoint name, not the target namespace or task queue.
+
+```
+Caller Namespace                          Handler Namespace
+┌──────────────┐     Nexus Endpoint     ┌──────────────────┐
+│ Caller       │─────────────────────▶  │ Handler Worker   │
+│ Workflow     │     (reverse proxy)    │ (polls target    │
+│              │◀─────────────────────  │  task queue)     │
+└──────────────┘                        └──────────────────┘
+```
```
dandavison, line 28: Do we have evidence that ASCII art is an effective way to communicate with an LLM? There's obviously reason to think that natural language might be more effective.

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Nexus Service
+
+A contract that defines a set of operations exposed through an endpoint. Services declare operation names and their input/output types. The contract is shared between caller and handler code.
+
+### Nexus Operation
+
+A single unit of work within a service. Operations follow a lifecycle that supports both:
+
+- **Synchronous operations** — Return immediately (within ~10 seconds). Suitable for quick lookups, validations, or triggering signals/updates on existing workflows.
```
dandavison, line 38: We should remember to add Standalone Activity here when the product release has got to a point where it's appropriate for it to be in the skill. Maybe we're already at that point?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+Nexus Endpoint
+    │  (routes to target namespace + task queue)
+    ▼
+Handler Worker
+    │  (polls for Nexus tasks, executes handler)
+    │
+    ├── Sync handler: returns result directly
+    │
+    └── Workflow run handler: starts workflow, Nexus tracks to completion
+```
```
dandavison, line 66: Ditto re diagrams

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Built-in Nexus Machinery
+
+The Nexus Machinery sits between caller and handler. It uses state-machine-based invocation and completion callbacks to provide:
```
dandavison, line 70: It's not clear to me what "state-machine-based invocation" means.

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+
+- **At-least-once execution** with automatic retries
+- **Circuit breaking** to prevent cascading failures
+- **Rate limiting** and **load balancing**
```
dandavison, line 74: Emphasis markup here not beneficial to an LLM?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+
+- **Synchronous operations** — Return immediately (within ~10 seconds). Suitable for quick lookups, validations, or triggering signals/updates on existing workflows.
+- **Asynchronous operations (Workflow Run)** — Start a workflow and track it to completion. Suitable for long-running work that benefits from Temporal's durable execution.
+
+### Nexus Registry
+
+The registry manages endpoint configuration. It is scoped to an Account in Temporal Cloud or to a Cluster for self-hosted deployments. Adding an endpoint to the registry deploys it so it is available at runtime.
+
+## How It Works
+
```
dandavison, line 46: Let's add here something like "Currently, Nexus operations may only be called from a Workflow. Calling from a Temporal client will come soon."

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+| `TRY_CANCEL` | Request cancellation and immediately report as cancelled |
+| `ABANDON` | Do not send a cancellation request |
+
+## Infrastructure Setup
+
+Nexus requires:
+1. **Separate namespaces** for caller and handler (or the same namespace for simpler setups)
+2. **A Nexus Endpoint** in the registry, routing to the handler's namespace and task queue
+3. **Handler workers** with Nexus service handlers registered
+4. **Caller workers** with workflows that create Nexus clients
```
dandavison, line 124: Consider dropping emph markup

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+| Asynchronous | `NexusOperationScheduled` → `NexusOperationStarted` → `NexusOperationCompleted` |
+
+## Multi-Level Calls
+
+Nexus supports chaining across multiple namespaces:
+
+```
+Namespace A          Namespace B          Namespace C
+Workflow A ──Nexus──▶ Workflow B ──Nexus──▶ Workflow C
+```
```
dandavison, line 93:
```suggestion
Nexus supports chaining across multiple namespaces: a workflow in namespace A can call a nexus operation that is handled in namespace B; that handler can in turn execute a workflow that calls a nexus operation that is handled in namespace C.
```

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+
+Each hop adds its own retry and fault isolation boundary.
+
+## Error Handling
+
+Nexus introduces two handler-side exception types and one caller-side type:
+
+- **Operation Error** — The operation itself failed. Non-retryable by default. Raised in handlers to indicate business-level failure.
+- **Handler Error** — A handler-level error with explicit retryability control. Use for transient infrastructure failures (retryable) or permanent handler issues (non-retryable).
```
dandavison, line 102: Could you reword these two so that they both start with "Raised in handlers"?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+- **Nexus Operation Error** — Raised in the caller workflow when a Nexus operation fails. Wraps the underlying cause, accessible via the error chain.
+
+Nexus operations share the same general error handling patterns as activities and child workflows — see `references/core/patterns.md` and `references/{your_language}/error-handling.md`.
```
dandavison, line 105: Would this be better worded as "Calling nexus operations shares the same..."?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+|-------------------|----------|
+| `WAIT_COMPLETED` (default) | Wait for the operation to fully complete after cancellation |
+| `WAIT_REQUESTED` | Wait for cancellation to be acknowledged by the handler |
+| `TRY_CANCEL` | Request cancellation and immediately report as cancelled |
+| `ABANDON` | Do not send a cancellation request |
+
+## Infrastructure Setup
+
+Nexus requires:
+1. **Separate namespaces** for caller and handler (or the same namespace for simpler setups)
```
dandavison, line 121: Let's clarify when same-namespace makes sense. Your table below clarifies it -- make the link or avoid duplicating the table content?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+| Same-namespace, same-team workflow composition | Maybe | Child workflows are simpler |
+| Single workflow calling activities | No | Use activities directly |
+| Simple parent-child workflow relationship | No | Use child workflows |
+
+## Best Practices
+
+1. **Keep service contracts shared** — Define in a module importable by both caller and handler code
+2. **Use business-meaningful workflow IDs** — In workflow run operations, derive IDs from business data for deduplication safety
+3. **Match operation type to duration** — Sync for <10s work, workflow run for anything longer
+4. **Co-locate handlers and workflows** — Register Nexus service handlers and their backing workflows on the same worker
```
dandavison, line 152: I believe current thinking is explaining briefly "why" to agents is good practice. What's the "why" here?

references/core/nexus.md
```diff
@@ -0,0 +1,154 @@
...
+5. **Design for at-least-once delivery** — Handlers should be idempotent, just like activities (see idempotency patterns in `references/core/patterns.md`)
+6. **Isolate failure domains** — Nexus provides namespace-level fault isolation; leverage it for independent deployment and scaling
```
dandavison, line 154: I don't think I know how to act upon this statement. What's it saying to do?

references/python/nexus.md
```diff
@@ -0,0 +1,280 @@
...
+        self, ctx: nexus.WorkflowRunOperationContext, input: MyInput
+    ) -> nexus.WorkflowHandle[MyOutput]:
+        return await ctx.start_workflow(
+            WorkflowStartedByNexusOperation.run,
+            input,
+            id=str(uuid.uuid4()),
+        )
+```
+
+**Note:** In production, prefer deterministic workflow IDs derived from business data for deduplication safety. `uuid.uuid4()` is shown for simplicity.
```
dandavison, line 92: Consider adapting the code sample to not use an UUID

references/python/nexus.md
```diff
@@ -0,0 +1,280 @@
...
+
+# Caller worker - runs in the caller namespace
+async def run_caller_worker():
+    client = await Client.connect("localhost:7233", namespace="my-caller-namespace")
+    worker = Worker(
+        client,
+        task_queue="my-caller-task-queue",
+        workflows=[CallerWorkflow],
+    )
+    await worker.run()
```
dandavison, line 141: Perhaps better to have theses in separate code blocks, maybe with `### Caller` and `### Handler` headings, since they are showing code essentially belonging to two different teams.

references/python/nexus.md
```diff
@@ -0,0 +1,280 @@
...
+            endpoint="my-nexus-endpoint-name",
+        )
+
+        # Execute and wait for result in one call
+        wf_result = await nexus_client.execute_operation(
+            MyNexusService.my_workflow_run_operation,
+            MyInput(name),
+        )
+
+        # Or start and handle separately
```
dandavison, line 170:
```suggestion
        # Or start and await handle separately
```

references/python/nexus.md
```diff
@@ -0,0 +1,280 @@
...
+        sync_handle = await nexus_client.start_operation(
+            MyNexusService.my_sync_operation,
+            MyInput(name),
+        )
+        sync_result = await sync_handle
```
dandavison, line 175: Would maybe be slightly more helpful to show `execute` example on the sync op and start+await handle example on the async op?

references/python/nexus.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+## Best Practices
+
+1. Keep service contracts in a shared module importable by both caller and handler code
+2. Use business-meaningful workflow IDs in workflow run operations for deduplication safety
+3. Use sync operations only for work that completes within 10 seconds; use workflow run operations for anything longer
+4. Register Nexus service handlers and their backing workflows on the same worker
+5. Use `execute_operation` when you just need the result; use `start_operation` when you need to cancel or manage the operation handle
+6. For multi-level Nexus calls (Workflow A → Nexus → Workflow B → Nexus → Workflow C), each hop adds its own retry and fault isolation
```
dandavison, line 280: What's the action item for point 6?

references/core/nexus.md
Quinn-With-Two-Ns: You mean when we have support to call Stand alone activities in a Nexus Operation, if so yep!

references/core/nexus.md
Quinn-With-Two-Ns: I saw some evidence from [blogs](https://www.dominguezpablo.com/blog/ascii-art-the-secret-weapon-for-better-ai-agent-performance) and posts that it helped AI, I have certainly had AI try to explain this to me with ascii art. That being said happy to remove if we aren't confident.

references/core/nexus.md
Quinn-With-Two-Ns: Other `Best Practices` sections don't explain "why" so I am not sure that is best practice or consistent with the rest of this skill, for example from error handling

```
## Best Practices

1. Use specific error types for different failure modes
2. Mark permanent failures as non-retryable
3. Configure appropriate retry policies
4. Log errors before re-raising
5. Use `ActivityError` to catch activity failures in workflows
6. Design code to be idempotent for safe retries (see more at `references/core/patterns.md`)
```

references/core/nexus.md
Quinn-With-Two-Ns: Emphasis markup is used all over this skill

references/core/nexus.md
Quinn-With-Two-Ns: We also use ascii art in lots of other places in this skill so it is consistent with the rest of the skill

dandavison: I agree LLM's can use ASCII art to explain things. I haven't researched the subject, but it really seems to me that ad-hoc ASCII art might be much less effective than text for conveying concepts to an LLM. Since these sorts of diagrams don't follow any formal syntax rules or have any widely-known conventions, it seems to me that tokenized ASCII-art diagram characters would key into the internal attention mechanisms of the trained model less effectively than would tokenized prose. I think @donald-pinckney agrees that it's worth re-assessing skill-wide. My hunch is that adding more diagrams is tech debt and that we'll replace them with prose.

PR #57, opened 2026-03-24
Add `@workflow.init` decorator to python.md Key Concepts

brianstrauch:
## What was changed
Added the new `@workflow.init` decorator to python.md.

## Why?
It was mentioned as being needed for Java [here](https://github.com/temporalio/skill-temporal-developer/pull/42#discussion_r2955189993), but it turns out to also be needed for Python. Adding for Java in the referenced PR.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/references/python/python.md
+++ b/references/python/python.md
@@ -98,7 +98,7 @@ if __name__ == "__main__":
 
 ### Workflow Definition
 - Use `@workflow.defn` decorator on class
-- Use `@workflow.init` on `__init__` to receive workflow input and initialize state before `@workflow.run` or any signal/update handler is invoked
+- Put any state initialization logic in the `__init__` of your workflow class to guarantee that it happens before signals/updates arrive. If your state initialization logic requires the workflow parameters, then add the `@workflow.init` decorator and parameters to your `__init__`.
 - Use `@workflow.run` on the entry point method
 - Must be async (`async def`)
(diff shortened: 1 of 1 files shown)


references/python/python.md
```diff
@@ -98,6 +98,7 @@ if __name__ == "__main__":
 
 ### Workflow Definition
 - Use `@workflow.defn` decorator on class
+- Use `@workflow.init` on `__init__` to run initialization code before `run` or any signal/update handler is invoked
```
brianstrauch:
```suggestion
- Use `@workflow.init` on `__init__` to receive workflow input and initialize state before `@workflow.run` or any signal/update handler is invoked
```

references/python/python.md
```diff
@@ -98,6 +98,7 @@ if __name__ == "__main__":
 
 ### Workflow Definition
 - Use `@workflow.defn` decorator on class
+- Use `@workflow.init` on `__init__` to receive workflow input and initialize state before `@workflow.run` or any signal/update handler is invoked
```
brianstrauch:
```suggestion
- Put any state initialization logic in the `__init__` of your workflow class to guarantee that it happens before signals/updates arrive. If your state initialization logic requires the workflow parameters, then add the `@workflow.init` decorator and parameters to your `__init__`.
```

PR #58, opened 2026-03-25
Add packaging workflow to release a public version and support Claude.ai Uploads

MasonEgger:
The skill works great as a Claude Code plugin where SKILL.md and references are auto-discovered, butusers may want to upload it to Claude.ai projects instead. You currently do this by uploaded a .zip file to your Claude.ai UI.

This adds a GitHub Actions workflow that packages the skill and references into a ZIP on every push to main, and creates a GitHub Release when the version in SKILL.md increases. Users can grab the ZIP from the release and upload it directly to a Claude.ai project without needing to clone the repo.

1. Any docs updates needed?
Maybe, if we want to call this out on our `with-ai` page to support this use case.


donald-pinckney (approved):
Logic looks good to me, I'll go ahead and merge, and test it out. Thanks!

PR #59, opened 2026-03-27
Update Go observability reference: replace outdated logur logging guidance with slog

ptdewey:
## What was changed

[references/go/observability.md](https://github.com/temporalio/skill-temporal-developer/blob/main/references/go/observability.md)

Replaces the outdated `logur` adapter guidance in the Go observability reference with `log.NewStructuredLogger` (and instructions for using bridges).

## Why?

Both slog (Go 1.21+) and the SDK's slog adapter (v1.25.0+) have been stable for a while now, so there's no reason to recommend pulling in logur (which has been archived since Jan 2025) as a dependency. The updated section documents `log.NewStructuredLogger` as the recommended approach, shows the slog bridge pattern for third-party loggers (with a zap example), and mentions direct `log.Logger` interface implementation as an alternative.

### DIFF
--- a/references/go/observability.md
+++ b/references/go/observability.md
@@ -28,6 +28,7 @@ func MyWorkflow(ctx workflow.Context, input string) (string, error) {
 ```
 
 The workflow logger automatically:
+
 - Suppresses duplicate logs during replay
 - Includes workflow context (workflow ID, run ID, etc.)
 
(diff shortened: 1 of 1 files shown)


donald-pinckney (approved):

donald-pinckney:
Thanks for the change, looks great!

PR #81, opened 2026-04-17
Auto-formatting: ran `mdformat --extensions frontmatter --number .`

donald-pinckney:
## What was changed
Ran a formatting pass over whole skill repo, via `pipx mdformat --extensions frontmatter --number .`

## Why?
To prepare for enforced autoformatting soon, to improve diffs, editing experience, and reduce manual formatting headache & churn.


donald-pinckney (commented):
Confirmed all diffs are correct formatting changes that are intended.

PR #107, opened 2026-05-07
fix: move SKILL.md into skills/temporal-developer/ for Claude Code discovery

imorrisonTwilio:
## Summary

- Claude Code discovers skills from `skills/<name>/SKILL.md`. With `SKILL.md` at the plugin root, the skill is not discovered and users get \"Unknown skill: temporal-developer\" errors.
- Moving to `skills/temporal-developer/SKILL.md` follows the Claude Code plugin convention and fixes skill discovery.

## Test plan

- [ ] Verify Claude Code can discover and invoke the `temporal-developer` skill after this change
- [ ] Confirm no other files need updating (README references, etc.)


donald-pinckney:
Hi, I think you would have skill discovery problems if you don't install the skill correctly using the install instructions in the README. We shouldn't need a `skills/` directory **in this repository**, you should be creating a `skills/` directory on your computer when installing the skill (or other skills).

Could you let me know how you installed the skill?

imorrisonTwilio:
Thanks for the feedback @donald-pinckney! You're right that this PR isn't needed.

The actual issue was that I was importing the wrong repo into our internal marketplace. `skill-temporal-developer` is the source/content repo designed for standalone cloning (where `SKILL.md` at the root is exactly correct). I should have been importing `temporalio/claude-temporal-plugin`, which is the purpose-built Claude Code plugin with the proper `skills/temporal-developer/SKILL.md` structure.

Once I switched the import to `claude-temporal-plugin`, skill discovery worked perfectly. Closing this PR.

PR #257, opened 2026-07-15
Document safe Workflow patch retirement

robzienert:
## Summary

Upstreams hardened versioning changes from our internal skills.

- clarify the generic patch lifecycle without assigning language-specific behavior to the core reference
- add a complete Go `workflow.GetVersion` retirement procedure covering recorded versions versus `DefaultVersion`, retention gates, marker placement/removal, and permanent `changeID` retirement
- align Python, TypeScript, Java, .NET, and Ruby patch-retirement guidance on retention, marker-absence ambiguity, and the difference between visibility population evidence and replay compatibility evidence
- correct Java visibility guidance for opt-in automatic upserts on 1.30+ and safe manual upserts of non-default versions

## Validation

- semantic content assertions across every language versioning reference
- balanced Markdown fence and local-link checks
- `git diff --check`
- fresh per-language Temporal-correctness and cross-document coherence review; final re-review clean

🤖 RZ's Aimee

### DIFF
--- a/references/core/versioning.md
+++ b/references/core/versioning.md
@@ -41,18 +41,20 @@ else:
 
 **Phase 1: Patch In**
 
-- Add both old and new code paths
-- New workflows take new path, old workflows take old path
+- Add both old and new code paths.
+- New executions take the new path; histories that predate the patch replay the old path.
 
 **Phase 2: Deprecate**
 
-- After all old workflows complete, remove old code
-- Keep deprecation marker for history compatibility
+- Wait until executions that can return the retired version have left retention before removing that branch or raising the minimum supported version.
+- Keep the Workflow's deprecation/version marker at the same deterministic location for history compatibility.
 
 **Phase 3: Remove**
 
-- After all deprecated workflows complete
-- Remove patch entirely, only new code remains
+- Removing the final marker is a separate operation with patching-API-specific rules beyond branch retirement.
+- Retire the patch/change identifier permanently when the patching API requires it; do not reuse an identifier whose first marker call was removed.
+
+See the language-specific versioning reference for exact marker, visibility, retention, and replay rules.
 
 ### When to Use
 
@@ -192,7 +194,7 @@ For long-running Workflows that cannot use Continue-as-New (e.g., compliance aud
 
 ## Best Practices
 
-1. **Check for open executions** before removing old code
+1. **Wait for affected executions to leave retention** before removing old code
 2. **Use descriptive patch IDs** (e.g., "add-fraud-check" not "patch-1")
 3. **Deploy incrementally**: patch → deprecate → remove
 4. **Test replay compatibility** before deploying changes
@@ -200,16 +202,11 @@ For long-running Workflows that cannot use Continue-as-New (e.g., compliance aud
 
 ## Finding Workflows by Version
 
+Different patching APIs encode `TemporalChangeVersion` values differently. A pre-patch execution has no marker for that patch, but marker absence can also mean an optional patch site has not executed. Follow the language-specific guide to query and classify the old-version population rather than guessing a marker value.
+
+Worker Deployment versions use a separate Search Attribute:
+
 ```bash
-# Find workflows with specific patch
-temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion = "add-fraud-check"'
-
-# Find pre-patch workflows
-temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion IS NULL'
-
-# Find workflows on specific worker version
 temporal workflow list --query \
   'TemporalWorkerDeploymentVersion = "my-service:v1.0.0"'
 ```
@@ -217,6 +214,8 @@ temporal workflow list --query \
 ## Common Mistakes
 
 1. **Removing old code too early** - Breaks replaying workflows
-2. **Not testing with replay** - Catches issues before production
-3. **Patching non-Command changes** - Unnecessary complexity
-4. **Forgetting to deprecate** - Accumulates dead code
+2. **Treating branch removal and final marker removal as identical changes** - Both can have retention gates, while final marker removal can also retire the identifier permanently
+3. **Treating marker absence as proof of an old branch** - Optional patch sites and other markers make absence ambiguous
+4. **Not testing with replay** - Replay catches compatibility failures before production
+5. **Patching non-Command changes** - Unnecessary complexity
+6. **Forgetting to deprecate** - Accumulates dead code
--- a/references/dotnet/versioning.md
+++ b/references/dotnet/versioning.md
@@ -68,7 +68,7 @@ public class OrderWorkflow
 
 **Step 2: Deprecate the Patch**
 
-Once all pre-patch Workflow Executions have completed:
+After all pre-patch Workflow Executions have left retention:
 
 ```csharp
 [Workflow]
@@ -92,7 +92,7 @@ public class OrderWorkflow
 
 **Step 3: Remove the Patch**
 
-After all workflows with the deprecated patch marker have completed, remove the `DeprecatePatch()` call entirely:
+In a later deployment, after the deprecation rollout is complete and the pre-patch Workflow Executions have left retention, remove the `DeprecatePatch()` call entirely. A deprecated marker is intentionally safe to omit during replay, so marker-bearing executions do not require another retention wait:
 
 ```csharp
 [Workflow]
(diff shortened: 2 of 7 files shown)


references/core/versioning.md
```diff
@@ -41,18 +41,20 @@ else:
...
-- New workflows take new path, old workflows take old path
+- Add both old and new code paths.
+- New executions take the new path; histories that predate the patch replay the old path.
 
 **Phase 2: Deprecate**
 
-- After all old workflows complete, remove old code
-- Keep deprecation marker for history compatibility
+- Wait until executions that can return the retired version have left retention before removing that branch or raising the minimum supported version.
+- Keep the SDK's deprecation/version marker at the same deterministic location for history compatibility.
```
robzienert: workflow, not "SDK"

references/core/versioning.md
```diff
@@ -200,23 +202,20 @@ For long-running Workflows that cannot use Continue-as-New (e.g., compliance aud
 
 ## Finding Workflows by Version
 
-```bash
-# Find workflows with specific patch
-temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion = "add-fraud-check"'
+The `TemporalChangeVersion` value format and retirement query are SDK-specific. For example, Go `GetVersion` and Java `getVersion` record `changeID-version`, while other patch APIs can use a bare patch ID. A pre-patch execution has no marker for that change ID, but marker absence can also mean an optional patch site has not executed. Follow the language-specific guide to query and classify the old-version population rather than guessing a marker value.
```
robzienert: Genericize this; `core` files aren't meant to have SDK-specific references.

references/go/versioning.md
robzienert, line 1: Do we need to provide the same treatment to the other language SDK `versioning.md`s?

starfleeth:
Thank you for this @robzienert ! Sorry about the delay. getting this reviewed ASAP.

references/core/versioning.md
```diff
@@ -41,18 +41,20 @@ else:
...
-- New workflows take new path, old workflows take old path
+- Add both old and new code paths.
+- New executions take the new path; histories that predate the patch replay the old path.
 
 **Phase 2: Deprecate**
 
-- After all old workflows complete, remove old code
-- Keep deprecation marker for history compatibility
+- Wait until executions that can return the retired version have left retention before removing that branch or raising the minimum supported version.
+- Keep the Workflow's deprecation/version marker at the same deterministic location for history compatibility.
```
Sushisource, line 50: It's worth mentioning you only need to wait until they've left _retention_ if you query them after they've completed, which is somewhat rare. If you know you don't do that, then you can remove the code when they've all completed, but before they've been archived/deleted.

references/core/versioning.md
```diff
@@ -41,18 +41,20 @@ else:
...
 
 **Phase 3: Remove**
 
-- After all deprecated workflows complete
-- Remove patch entirely, only new code remains
+- Removing the final marker is a separate operation with patching-API-specific rules beyond branch retirement.
+- Retire the patch/change identifier permanently when the patching API requires it; do not reuse an identifier whose first marker call was removed.
```
Sushisource, line 55: This seems unhelpful. It's basically saying "you can remove the marker when you should", whereas the old text tried to tell you when "should" is. Some combination of old/new would be appropriate here.

references/core/versioning.md
```diff
@@ -192,31 +194,28 @@ For long-running Workflows that cannot use Continue-as-New (e.g., compliance aud
 
 ## Best Practices
 
-1. **Check for open executions** before removing old code
+1. **Wait for affected executions to leave retention** before removing old code
```
Sushisource, line 197: Same comment about the difference between open & retention

references/core/versioning.md
```diff
@@ -192,31 +194,28 @@ For long-running Workflows that cannot use Continue-as-New (e.g., compliance aud
...
-```bash
-# Find workflows with specific patch
-temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion = "add-fraud-check"'
+Different patching APIs encode `TemporalChangeVersion` values differently. A pre-patch execution has no marker for that patch, but marker absence can also mean an optional patch site has not executed. Follow the language-specific guide to query and classify the old-version population rather than guessing a marker value.
 
-# Find pre-patch workflows
-temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion IS NULL'
+Worker Deployment versions use a separate Search Attribute:
```
Sushisource, line 207: This seems to have dropped a bunch of context about how to do queries with `TemporalChangeVersion`?

The previous paragraph refers to it without explaining what it is

references/dotnet/versioning.md
```diff
@@ -68,7 +68,7 @@ public class OrderWorkflow
 
 **Step 2: Deprecate the Patch**
 
-Once all pre-patch Workflow Executions have completed:
+After all pre-patch Workflow Executions have left retention:
```
Sushisource, line 71: Same on open/retention

references/dotnet/versioning.md
```diff
@@ -92,7 +92,7 @@ public class OrderWorkflow
 
 **Step 3: Remove the Patch**
 
-After all workflows with the deprecated patch marker have completed, remove the `DeprecatePatch()` call entirely:
+In a later deployment, after the deprecation rollout is complete and the pre-patch Workflow Executions have left retention, remove the `DeprecatePatch()` call entirely. A deprecated marker is intentionally safe to omit during replay, so marker-bearing executions do not require another retention wait:
```
Sushisource, line 95: This last sentence is hard to parse. I think it means something like "If a workflow's history contains a deprecated marker, it can safely be replayed by code without any of the old code or the patch call."

references/dotnet/versioning.md
```diff
@@ -121,11 +121,13 @@ Use List Filters to find workflows with specific patch versions:
...
 
-# Find running workflows without any patch (pre-patch versions)
+# Find retained workflows without any patch marker (simple pre-patch case)
 temporal workflow list --query \
-  'WorkflowType = "OrderWorkflow" AND ExecutionStatus = "Running" AND TemporalChangeVersion IS NULL'
+  'WorkflowType = "OrderWorkflow" AND TemporalChangeVersion IS NULL'
 ```
 
+`TemporalChangeVersion IS NULL` is reliable only when no other patch marker could make the attribute non-null. For multiple or optional patch sites, inspect all retained executions of the Workflow Type and classify the exact marker set and Event History. A zero running count identifies no immediate live blocker, but it does not prove that pre-patch histories have left retention. Replay selected histories to test compatibility; sampled replay does not prove the retained population is empty.
+
```
Sushisource, line 130:
```suggestion
`TemporalChangeVersion IS NULL` is reliable only when no other same-id patch marker could make the attribute non-null. For multiple or optional patch sites, inspect all retained executions of the Workflow Type and classify the exact marker set and Event History. A zero running count identifies no immediate live blocker, but it does not prove that pre-patch histories have left retention. Replay selected histories to test compatibility; sampled replay does not prove the retained population is empty.
```

references/go/versioning.md
```diff
@@ -15,9 +15,11 @@ v := workflow.GetVersion(ctx, "changeID", workflow.DefaultVersion, maxSupported)
 - `maxSupported`: current/newest version
 - Returns `maxSupported` for new executions; returns the recorded version on replay
 
-### Three-Step Lifecycle
+### Retirement Lifecycle
 
-**Step 1: Add GetVersion with both code paths**
+Treat `GetVersion` cleanup as replay-safety work, not ordinary dead-code removal. Removing an old branch and deleting the first marker call are separate changes with different gates.
```
Sushisource, line 20:
```suggestion
`GetVersion` requires care to remain replay safe. Removing an old branch and deleting the first marker call are separate changes with different gates.
```

references/go/versioning.md
```diff
@@ -32,26 +34,82 @@ if v == workflow.DefaultVersion {
 }
 ```
 
-For new executions, `GetVersion` returns `1` and records a marker. For replay of pre-change workflows (no marker), it returns `DefaultVersion` (`-1`).
+For new executions, `GetVersion` returns `1` and records a marker. For replay of pre-change workflows with no marker for `Step1`, it returns `workflow.DefaultVersion` (`-1`). **`DefaultVersion` is not version `0`**; do not search for `Step1-0` unless the code actually recorded version `0`.
+
+#### Step 2: Inventory and prove the old version has left retention
```
Sushisource, line 39: This section got really big in Go for some reason, but not the others (the smaller versions seem preferable. There's a lot of redundancy with the non-language-specific file).

Sushisource (commented):
There's quite a lot of llm-speak in here that I think is adding more tokens than value.

I think 90% of the value here would be captured by adding a line about the difference between open and retained workflows and what that means for you if you care about querying closed workflows. There's maybe also some good bits around using the search attribute correctly.

Most of the rest of it seems to be verbose-ifying information that's already there.

I think we could cut this down to a few lines added to the core file and that'd be better.

robzienert:
Ah sorry to waste your time on AI slop, but thanks for the feedback all the same. I'll take a second pass at this sometime this week.

PR #259, opened 2026-08-20
Fix Python reference guidance from issue #256

### LINKED ISSUE #256: Four nits in Python references and .memory/ notes that may be worth a look (verified against SDK 1.29.0)
### Overview, and thanks

I used the `temporal-developer` skill v0.5.0 (plugin 0.3.0) to review a production Python codebase, then asked the agent to verify each discovery against the installed SDK. I used a mixture of Fable and Opus 4.8 in this work. The skill is genuinely good, really helpful. I wish I'd found it sooner. It found real defects in our worker, and the remedies were easy to hand off to agents to fix. 

I found four cases where the skill gave bad advice. Each is against **temporalio 1.29.0**, Python 3.14. Upon inspection, `.memory/sdk-versions.json` records the references as verified against **1.22.0**, which may explain some drift. Also, all of the issues below were easily caught by validating agents and nothing incorrect made it into git, much less production.

The skill is intensive, really long, detailed, very helpful, and clearly made with great care. Each of these nits is low-hanging fruit that hopefully can be easily fixed and reduce the admittedly low amount of friction I encountered while running the tool. (I would not have bothered to file a bug report had the skill itself not encouraged it.) This is a spit-and-polish bug report, no more, no less.

#### 1. `.memory/correctness_checking/determinism-protection.md` suggests a nonexistent import

The note marks §7 "needs fixes" and says:

> Import path is INCORRECT: `temporalio.worker.workflow_sandbox` should be `temporalio.worker_sandbox`

This note appears to have it backwards. The reference file was already right:

```
>>> import temporalio.worker.workflow_sandbox   # works
>>> import temporalio.worker_sandbox
ModuleNotFoundError: No module named 'temporalio.worker_sandbox'
```

##### Possible fix

Deleting the note and marking §7 verified would prevent future confusion.

#### 2. `.memory/correctness_checking/observability.md` says `activity.logger` does not exist

§8 "Best Practices" is marked "needs fixes":

> **Best practice #1 is INCORRECT:** "`activity.logger` in activities" - `activity.logger` does NOT exist. Activities can use standard Python logging or print statements.

but it _does_ exist:

```
>>> from temporalio import activity
>>> activity.logger
<LoggerAdapter temporalio.activity (WARNING)>
```

And it's really useful, too! I definitely do not want to use standard Python logging when an Activity logger is available. This could mislead an agent to avoid a pretty handy part of the Temporal API. I'm glad that we already were, so our agent could say "wait a sec, hold on here."

##### Possible fix

§3 _in the same file_ verifies `activity.logger` as correct ("`activity.logger` API is correct ✓"), so the notes contradict each other within one file. The reference file's advice is right as written. Just removing "needs fixes" here could help.

#### 3. `@workflow.init` example is wrong about replay in `references/python/advanced-features.md`

The Workflow Init Decorator example carries:

```python
@workflow.init
def __init__(self, initial_value: str) -> None:
    # This runs only on first execution, not replay
```

But doesn't `__init__` run on replay too? Claude sure thinks so: "A replaying workflow is a freshly constructed instance: `_WorkflowInstanceImpl._instantiate_workflow_object()` is invoked from `_run_once()` whenever `not self._object and not self._deleting` (`_workflow_instance.py:2539`), with no replay check." The comment invites exactly the bug the surrounding section is trying to prevent, namely, putting something in `__init__` believing it is replay-exempt.

##### Possible fix

`.memory/correctness_checking/advanced-features.md` already flags this section "needs fixes" and says the claim "needs verification". I'd just delete the comment.

#### 4. `references/core/versioning.md` conflates Worker Deployments with the deprecated Build-ID API

Line 11 describes Worker Versioning as "Deployment-level control with Build IDs", and the doc uses Build-ID framing throughout. But in 1.29.0 these are distinct, and what's more, `Client.get_worker_build_id_compatibility` and `update_worker_build_id_compatibility` both carry 

```
.. deprecated::
    Legacy API, see the docs above for new usage
```

The current API is `WorkerDeploymentConfig` + `WorkerDeploymentVersion`, where Build ID survives only as a field of a deployment version.

Temporal's own note at `.memory/correctness_checking/versioning.md` says the Python "Choosing a Strategy" section is missing a warning "The pre-2025 Worker Versioning API (Build ID-based versioning) is being removed in March 2026." The TypeScript reference already carries exactly that note (`typescript/versioning.md:151`), and its QA entry is marked FIXED; `core/` and `python/` just seem to have missed out.

##### 100% this part is written by Claude. Since the target audience is agents, that seemed only fair.

**Surface the `.memory/` status in `SKILL.md`.** The correctness notes are excellent and I would have read them first had I known they existed — but nothing in `SKILL.md` mentions them, so I reviewed a codebase against references whose own summary table marks Python `partial` for 12 of 13 files. A line in `SKILL.md` saying "the Python references are partially verified; see `.memory/correctness_checking/` before relying on a claim" would set the right prior. Same for `sdk-versions.json`: knowing the refs were checked at 1.22.0 while I ran 1.29.0 is exactly the caveat a reviewer needs up front.

**Consider whether `.memory/` should ship at all.** If it is internal QA rather than user-facing, shipping it in the plugin cache invites agents to read it as authoritative — which is how I hit 1 and 2. Either exclude it, or state its status in `SKILL.md`.

##### Again, these words are Claude's

For balance, the skill was right and useful about: the sandbox/passthrough model, `workflow_failure_exception_types` for tests (which closed a real hang risk for us), the 100-slot activity default, `workflow.logger` replay-safety, payload limits, Task Queue Fairness for multi-tenant starvation, and the "check for open executions" quiet-window practice. I verified the SDK-side ones against 1.29.0; the server-side claims (payload limits, Task Queue Fairness) I took on trust. It also correctly identified that our `Runtime.set_default` usage diverged from the canonical `Client.connect(runtime=...)` — a genuinely good catch buried in a `needs fixes` note.

brianstrauch:
## Summary

- clarify that `@workflow.init` initialization runs during replay
- distinguish current Worker Deployment Versioning from deprecated Build ID compatibility-set APIs
- update the Python Worker Deployment imports to the current public API
- note that the problematic `.memory/` QA files are already absent from the packaged repository

## Validation

- `git diff --check`
- cross-checked the versioning terminology and Python imports against current Temporal documentation and SDK sources

Closes #256

### DIFF
--- a/references/python/versioning.md
+++ b/references/python/versioning.md
@@ -183,6 +183,9 @@ temporal workflow list --query 'WorkflowType = "PizzaWorkflow" AND ExecutionStat
 
 Worker Versioning manages versions at the deployment level, allowing multiple Worker versions to run simultaneously.
 
+> [!IMPORTANT]
+> Use the Worker Deployment APIs described below. The older Build ID-based APIs, including `Client.get_worker_build_id_compatibility()` and `Client.update_worker_build_id_compatibility()`, manage legacy compatibility sets and are deprecated; they are not the same as Worker Deployment Versioning.
+
 ### Key Concepts
 
 **Worker Deployment**: A logical service grouping similar Workers together (e.g., "loan-processor"). All versions of your code live under this umbrella.
@@ -192,11 +195,8 @@ Worker Versioning manages versions at the deployment level, allowing multiple Wo
 ### Configuring Workers for Versioning
 
 ```python
-from temporalio.worker import Worker
-from temporalio.worker.deployment_config import (
-    WorkerDeploymentConfig,
-    WorkerDeploymentVersion,
-)
+from temporalio.common import WorkerDeploymentVersion
+from temporalio.worker import Worker, WorkerDeploymentConfig
 
 worker = Worker(
     client,
@@ -217,7 +217,7 @@ worker = Worker(
 
 - `use_worker_versioning`: Enables Worker Versioning
 - `version`: Identifies the Worker Deployment Version (deployment name + build ID)
-- Build ID: Typically a git commit hash, version number, or timestamp
+- `build_id`: The code-version component of a Worker Deployment Version, typically a git commit hash, version number, or timestamp
 
 ### PINNED vs AUTO_UPGRADE Behaviors
 
--- a/references/python/versioning.md
+++ b/references/python/versioning.md
@@ -184,7 +184,7 @@ temporal workflow list --query 'WorkflowType = "PizzaWorkflow" AND ExecutionStat
 Worker Versioning manages versions at the deployment level, allowing multiple Worker versions to run simultaneously.
 
 > [!IMPORTANT]
-> Use the Worker Deployment APIs described below. The older Build ID-based APIs, including `Client.get_worker_build_id_compatibility()` and `Client.update_worker_build_id_compatibility()`, manage legacy compatibility sets and are deprecated; they are not the same as Worker Deployment Versioning.
+> Use the Worker Deployment APIs described below. The older Build ID-based APIs, including `Client.get_worker_build_id_compatibility()` and `Client.update_worker_build_id_compatibility()`, manage legacy compatibility sets and are deprecated.
 
 ### Key Concepts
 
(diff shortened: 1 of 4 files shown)


references/typescript/versioning.md
```diff
@@ -148,7 +148,7 @@ After all V1 executions complete, remove the old Workflow function.
 
 Worker Versioning allows multiple Worker versions to run simultaneously, routing Workflows to specific versions without code-level patching. Workflows are pinned to the Worker Deployment Version they started on.
 
-> **Note:** Worker Versioning is currently in Public Preview. The legacy Worker Versioning API (before 2025) will be removed from Temporal Server in March 2026.
+> **Note:** Worker Versioning is currently in Public Preview. The legacy Worker Versioning API is deprecated.
```
brianstrauch:
```suggestion
```

references/python/versioning.md
```diff
@@ -183,6 +183,9 @@ temporal workflow list --query 'WorkflowType = "PizzaWorkflow" AND ExecutionStat
 
 Worker Versioning manages versions at the deployment level, allowing multiple Worker versions to run simultaneously.
 
+> [!IMPORTANT]
+> Use the Worker Deployment APIs described below. The older Build ID-based APIs, including `Client.get_worker_build_id_compatibility()` and `Client.update_worker_build_id_compatibility()`, manage legacy compatibility sets and are deprecated.
```
brianstrauch:
```suggestion
> Use the Worker Deployment APIs described below. The older Build ID-based APIs manage legacy compatibility sets and are deprecated.
```

